PR #1, opened 2016-08-24
Adding support for Apple Xcode, and run conan install on first configure.

Xaltotun:
This script still requires an explicit dependency from your target to the conan-install target.   I plan to fix this in a future revision.

### DIFF
--- a/conan.cmake
+++ b/conan.cmake
@@ -1,5 +1,13 @@
 include(CMakeParseArguments)
 
+set(POLICY_CHECK "")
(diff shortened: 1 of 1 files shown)


memsharded:
Hi @Xaltotun,

Sorry for this waiting so long, this feature was very delayed, haven't resumed working on it till now. We have moved the repo to the official conan-io project, added tests to it, a Readme, documented in the conan docs, and waiting for users feedback.

I simplified the targets, and right now just using execute_process, which is much simpler. 
Thanks for your contribution!

PR #4, opened 2017-01-07
Fix incorrect SETTINGS_STR variable use

GatorQue:
The SETTINGS_STR variable should be _SETTINGS ${_SETTINGS} for GCC and AppleClang sections of conan_cmake_settings function.


memsharded:
Yes, seems a mistake while copying & pasting, thanks for fixing this!

PR #5, opened 2017-01-07
use prefered CMake variable MSVC_VERSION for visual studio version check

GatorQue:
CMake provides the MSVC_VERSION variable for detecting the Visual Studio version. Use it instead of CMAKE_GENERATOR variable.


memsharded:
Nice job, thanks very much!

PR #6, opened 2017-01-12
Add detection of Visual Studio runtime from CMake environment

sixten-hilborn:
Currently only MD is hardcoded, for debug MDd should be used by default, and any other setting in CMake should override this.


memsharded:
Cool CMake-fu here, thanks a lot!

PR #8, opened 2017-01-18
Auto-detect GCC libcxx flag

sixten-hilborn:
This commit fixes the absence of libstdc++11 for GCC

### DIFF
--- a/conan.cmake
+++ b/conan.cmake
@@ -26,7 +26,8 @@ function(conan_cmake_settings result)
     string(REPLACE "." ";" VERSION_LIST ${CMAKE_CXX_COMPILER_VERSION})
     list(GET VERSION_LIST 0 MAJOR)
     list(GET VERSION_LIST 1 MINOR)
-    set(_SETTINGS ${_SETTINGS} -s compiler=gcc -s compiler.version=${MAJOR}.${MINOR} -s compiler.libcxx=libstdc++)
+    conan_cmake_detect_gnu_libcxx(_LIBCXX)
+    set(_SETTINGS ${_SETTINGS} -s compiler=gcc -s compiler.version=${MAJOR}.${MINOR} -s compiler.libcxx=${_LIBCXX})
   elseif (${CMAKE_CXX_COMPILER_ID} STREQUAL AppleClang)
       # using AppleClang
       string(REPLACE "." ";" VERSION_LIST ${CMAKE_CXX_COMPILER_VERSION})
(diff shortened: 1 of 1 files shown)


conan.cmake
```diff
@@ -83,6 +84,22 @@ function(conan_cmake_settings result)
...
 
+function(conan_cmake_detect_gnu_libcxx result)
+    get_directory_property(defines DIRECTORY ${CMAKE_SOURCE_DIR} COMPILE_DEFINITIONS)
+    foreach(define ${defines})
+        if(define STREQUAL "_GLIBCXX_USE_CXX11_ABI=0")
+            set(${result} libstdc++ PARENT_SCOPE)
+            return()
+        endif()
+    endforeach()
+    if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS "5.1")
```
memsharded: The difficult part of this, and the reason why libcxx is explicitely modeled as a setting, parallel to the version one, is that this is not always true: A typical installation of gcc 5.3/5.4 in a (mainstream) Ubuntu 14, will not install libstdc++11, but libstdc++ (due to system/installed apps incompatibilities).

I guess that it is enough defining ``cmake -D_GLIBCXX_USE_CXX11_ABI=0`` in the command line, is it? Not very sure about what is the behavior of ``get_directory_property`` for the defines

sixten-hilborn: Hm, that's unfortunate that gcc 5.1+ might not have libstdc++11 available even though it's supposed to...

The usage of `get_directory_property` checks for definitions (added via `add_definitions`) to see if `_GLIBCXX_USE_CXX11_ABI` has been defined to 0 (which is the way to tell GCC to use libstdc++ instead of libstdc++11).

memsharded: Yes, really unfortunate, but the real pain is that it is very common, for example in **travis-ci** you have old ubuntu machines, and a typical gcc upgrade to 5.4 won't get a libstdc++11.

memsharded: In any case, your approach is much better than the existing, so lets merge it, thanks very much!

memsharded:
I have been playing a bit with this, after merging, and found one issue. I am trying to provide users of gcc 5.1+ and libstdc++ a way to specify that they want to use the old libstdc++ instead of the new one. For me a convenient/intuitive way would be to specify it as:

```bash
$ cmake . -G "Unix Makefiles" -D_GLIBCXX_USE_CXX11_ABI=0 -DCMAKE_BUILD_TYPE=Release
```

Unfortunately, this doesn't work, because the check of ``_GLIBCXX_USE_CXX11_ABI`` is done over directory COMPILE_DEFINITIONS, and command line definitions are disregarded.

Any suggestion to handle this case? Thanks!

PR #9, opened 2017-01-18
Add support for the cmake_multi generator

sixten-hilborn:
Use Conan's new cmake_multi generator if a CMake generator with multi-configuration support is used (like Visual Studio) and the user does not explicitly state a CMAKE_BUILD_TYPE.


memsharded:
Looks very good, thanks very much!
Just merged, I'll try to add a couple of tests for it.

Issue #10, opened 2017-01-22
Using conan.cmake in packages/libraries


sixten-hilborn:
Hi,

I'm developing a library that I want to turn into a conan package. It has some dependencies so I want to import them into CMake via `conan.cmake` while working on the library. However with this setup `conan_cmake_run` will do another call to `conan install` when installing the exported package.

I propose that `conan_cmake_run` will merely include `conanbuildinfo.cmake` and call `conan_basic_setup` after the package has been exported. Does conan set some flag when building a package that is easy to check, like `CONAN_INSTALLING` or `CONAN_EXPORTED`?

One limitation when using `conan.cmake` in a package/library is that a conanfile.py is needed so it's not possible to use the `REQUIRES` arguments to `conan_cmake_run`.

PR #13, opened 2017-01-22
Support conan.cmake in exported packages

sixten-hilborn:
See #10.


conan.cmake
```diff
@@ -197,19 +197,30 @@ endmacro()
 
 macro(conan_cmake_run)
     parse_arguments(${ARGV})
-    conan_cmake_generate_conanfile(${ARGV})
-    if(CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
+
+    # TODO: Better detection
+    if(CONAN_EXPORTED OR CONAN_COMPILER)
```
memsharded: I am not sure if I'd use CONAN_COMPILER for detection, some users might be feeding it from the command line to cmake. CONAN_EXPORTED would do the task, wouldn't it? Is it necessary to set it to ``ON`` then?

tests.py
```diff
@@ -169,3 +169,42 @@ class Pkg(ConanFile):
       run("cmake .. -DCMAKE_BUILD_TYPE=Release")
       run("cmake --build . --config Release")
       run("bin\main")
+
+    def test_exported_package(self):
+      content = """cmake_minimum_required(VERSION 2.8)
+project(conan_wrapper CXX)
+
+set(CONAN_EXPORTED ON)
```
memsharded: I guess we should remember to remove this ``CONAN__EXPORTED ON`` line, when the conan codebase already sets it automatically.

conan.cmake
sixten-hilborn: Agree, this is a temporary workaround before CONAN_EXPORTED was defined by conan, but it's supposed to work with both 0.18 and 0.19.

tests.py
sixten-hilborn: That's true, it's not needed from the next conan release.

PR #16, opened 2017-02-01
handle CMAKE_SYSTEM_NAME as conan -s os=System

cguentherTUChemnitz:
Common cmake toolchain [files](https://github.com/jcupitt/build-win64/blob/master/8.2/Toolchain-x86_64-w64-mingw32.cmake) sets the CMAKE_SYSTEM_NAME. This one can be used as the conan os setting. This os detection is interesting for cross compiling workflows like mingw linux to windows compilation.

### DIFF
--- a/conan.cmake
+++ b/conan.cmake
@@ -26,11 +26,15 @@ function(conan_cmake_settings result)
   endif()
 
   #handle -s os setting
-  set(CONAN_SUPPORTED_PLATFORMS Windows Linux Macos Android iOS FreeBSD)
-  if( CMAKE_SYSTEM_NAME IN_LIST CONAN_SUPPORTED_PLATFORMS )
-    set(_SETTINGS ${_SETTINGS} -s os=${CMAKE_SYSTEM_NAME})
-  else()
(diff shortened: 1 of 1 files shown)


memsharded:
But what happens if using plain CMake, and CMAKE_SYSTEM_NAME  is not defined? (I think it is usually not defined, unless cross-building). Your PR will raise a FATAL_ERROR. I think that if the variable is not defined, no check should be done, and the default (host) ``os`` setting should be valid.

Does it make sense?

cguentherTUChemnitz:
I agree. Your proposal would not break systems without defined CMAKE_SYSTEM_NAME. I use currently the QtCreator cmake integration. This one seems to set the CMAKE_SYSTEM_NAME correctly from itself, but you are right that this can not be assumed for other use-cases.

memsharded:
The included tests are failing, did you run them? I am checking right now.

memsharded:
I have replaced the IN_LIST for old-style list find: https://github.com/conan-io/cmake-conan/commit/9e352d10748f3ac92167539cea4aba297bdfe136

The issue were the CMake policies, and I am using cmake 3.5, but if you don't set explicitely the policy it will fail. I think the old-style will be easier for more people, what do you think?

cguentherTUChemnitz:
Sorry i did not run them. I have no preference to the CMake version. I think also that the older version would be more compatible to older setups. So i agree to your proposed changeset.

PR #17, opened 2017-02-08
Recognize MacOS

alandefreitas:
I slightly changed the script to recognize Darwin (as CMake calls the MacOS) as Macos (as Conan calls the MacOS).


memsharded:
Thanks for fixing this!

PR #20, opened 2017-02-19
Re-run cmake when conanfile is changed (#18)

### LINKED ISSUE #18: Re-run cmake configure step when conanfile is updated
Hi!
It would be convenient if CMake kept track of the conanfile, and re-ran the cmake scripts when building if conanfile has been changed. One "workaround" at the moment is to touch CMakeLists.txt if a requirement is updated in the conanfile, and then build.

sixten-hilborn:
Fixes #18 by adding the conanfile to CMake's generate.stamp.depend file.

### DIFF
--- a/conan.cmake
+++ b/conan.cmake
@@ -186,14 +186,22 @@ function(conan_cmake_install)
 endfunction()
 
 
+function(conan_cmake_setup_conanfile)
+  parse_arguments(${ARGV})
+  if(ARGUMENTS_CONANFILE)
+    # configure_file will make sure cmake re-runs when conanfile is updated
+    configure_file(${ARGUMENTS_CONANFILE} ${ARGUMENTS_CONANFILE}.junk)
+    file(REMOVE ${CMAKE_CURRENT_BINARY_DIR}/${ARGUMENTS_CONANFILE}.junk)
+  else()
+    conan_cmake_generate_conanfile(${ARGV})
+  endif()
+endfunction()
+
 function(conan_cmake_generate_conanfile)
   # Generate, writing in disk a conanfile.txt with the requires, options, and imports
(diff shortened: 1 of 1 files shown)


conan.cmake
```diff
@@ -186,14 +186,22 @@ function(conan_cmake_install)
 endfunction()
 
 
+function(conan_cmake_setup_conanfile)
+  parse_arguments(${ARGV})
+  if(ARGUMENTS_CONANFILE)
+    # configure_file will make sure cmake re-runs when conanfile is updated
+    configure_file(${ARGUMENTS_CONANFILE} ${ARGUMENTS_CONANFILE}.junk)
```
memsharded: Hi Sixten. This functionality seems interesting but a couple of issues:
- Tests are broken, could you please check them?
- I don't fully understand what are you trying to do here. ``configure_file`` could be useful, but I am not sure how will you get cmake re-ran, or why do you just remove the .junk file after the command. Most likely some cmake-fu I ignore, could you please explain a little? Thanks!

sixten-hilborn: Hi, this is indeed some cmake-fu. From the documentation https://cmake.org/cmake/help/v3.0/command/configure_file.html:

> If the \<input\> file is modified the build system will re-run CMake to re-configure the file and generate the build system again.

So CMake will keep track of the input file if it changes, like it does with `include`'d .cmake files, by storing it in `${CMAKE_BINARY_DIR}/CMakeFiles/generate.stamp.depend`. However `configure_file` also requires an output file to be specified, but I remove it immediately to avoid confusion since we're not going to use it.

memsharded: Cool, thanks for the explanations, merged! :)

PR #22, opened 2017-02-27
Use CMAKE_CURRENT_BINARY_DIR instead of CMAKE_BINARY_DIR

g-s-m:


memsharded:
Looks good, I think it can be merged, but I would like the feedback of @sixten-hilborn if possible, please.

The use case would be that you are calling it from different sub-projects/sub-folder, but they are not using the same dependency graph, but each one has its own dependency graph, with its own requirements and configuration, right?

g-s-m:
yes, that's right. I have a project and another project is included as a git submodule. Both project has its own dependency graph. I include submodule in  build with cmake add_subdirectory command. With these changes conanbuild*.cmake files will be generated for each project in its own build directory

memsharded:
Thanks very much for the contribution, merged and created a new release: https://github.com/conan-io/cmake-conan/releases

PR #26, opened 2017-04-04
Support MSVC toolset

KindDragon:


memsharded:
What would be the purpose of this PR?
I have ran the included test suite forcing to use the VS 15 2017 generator, and it works fine.

KindDragon:
It doesn't work well if you run cmake for example with parameters `-G "Visual Studio 15 2017 Win64" -T v140` (with toolset from VS2015)

memsharded:
Ok, thanks. Good contribution, thank you!

PR #38, opened 2017-07-05
Allow passing mutiple `--build` arguments to `conan install`

r4zzz4k:
`conan install` documentation [mentions](http://conanio.readthedocs.io/en/latest/reference/commands/install.html#build-options) that it's possible to pass multiple `--build` arguments with patterns describing packages. This pull request allows one to use this feature from _cmake-conan_.

For example, this invocation:
```cmake
conan_cmake_run(
    REQUIRES ...
    OPTIONS ...
    ...
    BUILD gtest spdlog
)
```
will lead to issuing following command: ` conan install ... --build=gtest --build=spdlog`, which will build mentioned dependencies from source, but using prebuilt variants of others.


memsharded:
Thanks very much for fixing this!

r4zzz4k:
Don't mention it. Thank you for this useful and convenient tool!

PR #39, opened 2017-07-12
Pass MinSizeRel and RelWithDebInfo CMAKE_BUILD_TYPE as Release to Conan's build_type

r4zzz4k:
If we invoke CMake generator stage passing specific build type, which causes to use `cmake` Conan generator, `CMAKE_BUILD_TYPE` is passed as-is to conan `-s build_type` argument. So if we try to run
`cmake -DCMAKE_BUILD_TYPE="RelWithDebInfo" ...`, then `conan -s build_type=RelWithDebInfo` is invoked, which causes it to fail mentioning it knows only `Debug` and `Release` build types. This pull request converts `CMAKE_BUILD_TYPE`s `MinSizeRel` and `RelWithDebInfo` to `-s build_type=Release` so that dependencies are built by Conan properly.


r4zzz4k:
Oops, after removing C:\.conan and ~\.conan and doing clean build I'm having issues. Investigating.

r4zzz4k:
Sorry, my bad -- forgot to pass `--config` to CMake on build stage and for some reason got Debug build even though I set `CMAKE_BUILD_TYPE` during configuration stage explicitly. This pull request should be fine.

memsharded:
Good, thanks very much for this contribution!

PR #45, opened 2017-09-19
Change target platform detection under MSVC.

### LINKED ISSUE #44: Under Visual Studio 15.3.4 lack of $PLATFORM environment variable causes generation to fail
This seem to happen only under VS IDE. Standalone CMake 3.9.0 with same VS15 toolchain, same machine, same project - didn't have this problem. Maybe that effect of VS using Ninja internally. The workaround is to create this variable, but that causes problems on different occasions.

```
Severity	Code	Description	Project	File	Line	Suppression State
Error		CMake Error at C:/Users/#####/CMakeBuilds/316c7597-0638-b33f-a43a-f74b1dfab3d1/build/x86-Debug/conan.cmake:115 (if):
  if given arguments:

    "STREQUAL" "X64"

  Unknown arguments specified		C:/Users/#####/CMakeBuilds/316c7597-0638-b33f-a43a-f74b1dfab3d1/build/x86-Debug/conan.cmake	115	
```

unjello:
`$env:Platform` may be undefined. This caused Visual Studio 2017 to fail
when CMake project was opened with an IDE. Target architecture can be also
infered from `_MSVC_CXX_ARCHITECTURE_FAMILY` and/or `_MSVC_C_ARCHITECTURE_FAMILY`.

Closes #44


memsharded:
This looks good, thanks very much. Sorry for taking so long to process, for some reason we didn't get then notification and failed to notice it. Feel free to ping us anytime if we don't respond. Thanks again!

PR #46, opened 2017-10-17
Add suport for using profiles.

bobeff:
Add DEBUG_PROFILE, RELEASE_PROFILE and PROFILE arguments to conan_cmake_run macro. When supplyed:
  * DEBUG_PROFILE - overrides settings for DEBUG builds
  * RELEASE_PROFILE - overrides settings for RELEASE builds
  * PROFILE - overrides settings for all build types.


memsharded:
Thanks for contributing this!

However, I am not sure I fully understand the approach. Why inject a profile, if it could be incompatible with the current settings?

When cmake is running, it already identified an is using a certain compiler, version, architecture, etc. Using a profile can only conflict with this, it will certainly fail to install dependencies for a different configuration than the given one.

Can you please explain a bit more how do you deal with this issue? Thanks!

bobeff:
I left compatibility of the profile  to be ensured by the user.

Support for profiles is crucial for out projects not only for specifying of required options which can be different for different platforms, but because some of the settings also are not determined correctly.

For example we build out projects on Ubuntu Linux with `g++ ` version `7.2`. We are using C++17 standard and the script determines the standard library as `libstdc++11` but there is no such thing for Ubuntu. There is only `libstdc++` and it supports the last version of standard implemented by the compiler.

There is possibility for mistakes if the user specifies a wrong profile, but this is better compared to missing any profiles support.

sixten-hilborn:
Hi,

Since gcc 5.1 there's been modifications to std::string and some other stuff in stdlibc++ to conform to C++11 which breaks binary compatibility, so newer versions of stdlibc++ actually contains two implementations of std::string. stdlibc++11 is Conan's terminology for using the newer implementation.

I like profiles for things like `build_requires`, but I'm not sure about the debug/release profiles together with `conan.cmake`.
Would it be useful to have an argument to be able to disable settings detection and/or automatic conan innovation?

bobeff:
> Since gcc 5.1 there's been modifications to std::string and some other stuff in stdlibc++ to conform to C++11 which breaks binary compatibility, so newer versions of stdlibc++ actually contains two implementations of std::string. stdlibc++11 is Conan's terminology for using the newer implementation.

Thanks for clarification. 

> I like profiles for things like build_requires, but I'm not sure about the debug/release profiles together with conan.cmake.

On windows CMake Visual Studio generator is multi configuration and this required both debug and release profile because usually they are different. At least `build_type` setting is different. In some cases options also can be different. For example using shared version of some library in Debug builds for better build time, but static version in Release builds for not redistributing dll files.

> Would it be useful to have an argument to be able to disable settings detection and/or automatic conan innovation?

In my implementation setting profile manually overrides settings detected by Conan.

memsharded:
I think @bobeff explanation makes sense. 

It is a feature that enable usage of build_requires, and other things like options that might be managed better in profiles. 

It can also fix some mis-detection of settings

It is clearly opt-in, so users not using it, won't be affected at all. So I would go for merging it.

@sixten-hilborn yes, but if you disable the settings detection, then you need to provide the settings yourself, or it won't work. And for that, the best is providing a profile. Wdyt?

sixten-hilborn:
Yes it sounds good, and the implementation looks fine to me. :)

PR #48, opened 2017-11-16
Added support for OPTIONS when using the CONANFILE parameter.

lbakman:
Hi

I ran into a situation today where I needed to provide options when calling conan from CMake. This works if using the REQUIRES parameter in conan_cmake_run but not if you are using the CONANFILE parameter.

This pull request includes the changes I have made to conan.cmake to support options in the above scenario.

Regards 
Lau


memsharded:
Yes, good catch. It was forgotten, because you can specify options in the conanfile, but yes, they should be used if passed in CMake. Many thanks!

PR #49, opened 2017-11-23
Removing override for RelWithDebInfo and MinSizeRel

parsifal-47:
Hi,
I understand that it was probably made on purpose, but if I try to build my project with debug info, all deps are built with "Release".

if you need it for reason, we'll try to find another way around.

Thanks


memsharded:
Could you please clarify a bit further what the problem is? If you try to build in Release mode (with Debug info, but Release mode), dependencies will be installed as "Release" too, which works. 
A not exactly related, but relevant issue here is that the recommended flow is not to build the dependencies when doing a ``conan install``. The best flow is if those dependencies are already built, and only the binaries have to be retrieved.

I think you might want to discuss with @r4zzz4k, which introduced these lines

Also, the CI has to be fixed. I will try to do it now.

parsifal-47:
the problem is: I'm building in `RelWithDebInfo`, which is not completely `Release` mode and I have my deps pre-built as binaries in `RelWithDebInfo` as well, but `RelWithDebInfo` got overriden with `Release` so my deps got rebuit and I have no information for stack-traces for example.

It works, but I loose symbols for deps and it invokes my build once again, technically, I'm not able to use any prebuilt binaries in `RelWithDebInfo`

parsifal-47:
There is no way to use deps in `RelWithDebInfo` mode from cmake.
with `conan install` I could do that with `-s build_type=RelWithDebInfo`, but not from cmake, that is the problem

memsharded:
Ok, makes sense, thanks! Could you please @r4zzz4k comment about this?

Also, I have fixed CI, to comply with latest conan 0.29.1, tests should be green now.

r4zzz4k:
Looks like it's okay to merge. In my PR #39 I mention that `conan` doesn't know `RelWithDebInfo` target, and it seems that this was implemented in https://github.com/conan-io/conan/pull/1325. So workaround in question can probably be removed at this point.

memsharded:
Ok, thanks @r4zzz4k for checking this, and thanks @parsifal-47 for the contribution!

PR #54, opened 2017-12-25
fix  conan install  command in accordance with 1.0.0b2

Artalus:
1.0.0-beta changed `conan install` syntax to use `path/to/conanfile.txt` instead of `-f=path/to/conanfile.txt`; this small fix changes invokation of `conan install` during `conan_cmake_install` function.

### DIFF
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -19,6 +19,6 @@ if [[ "$(uname -s)" == 'Darwin' ]]; then
     pyenv activate conan
 fi
 
-pip install conan
+pip install conan==1.0.0b2
 conan user
 pip install nose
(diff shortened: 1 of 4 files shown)


memsharded:
Thanks very much, great job!

Artalus:
Take note that last commit changes `pip install conan` to `pip install conan==1.0.0b2` in tests scripts. This better be rolled back once Conan 1.0 leaves beta and `pip install` will pull the latest release.

memsharded:
Yes, absolutely. This is why I haven't done a release yet with these commits. They might break, but if someone complains, I will tell to use the latest release, and we will release on due time for 1.0. Thanks!

PR #55, opened 2018-01-04
Use gcc major version for package installation.

### LINKED ISSUE #53: How to use install packages build for gcc major versions?
`conan-package-tools` recently introduced gcc major versions (for gcc  version >= 5). How to use these? `cmake-conan` detects e.g. `compiler.version=5.4` but I want to use a package which was build with the `conangcc5`-docker image. Has `cmake-conan` to be adapted?

bilke:
Closes #53.

Not backwards-compatible!


memsharded:
Thanks for the contribution!
Even if not backward compatible:
- Tagged releases have been done in this repo for a while. People not moving forward should use tagged releases
- Conan 0.X is going to be deprecated in a matter of few days.

So I think it can be merged safely.

parsifal-47:
Hi, is it consistent now?
I'm getting this error message and don't really understand why you cut minor part of gcc version:
```
ERROR: Invalid setting '7' is not a valid 'settings.compiler.version' value.
Possible values are ['4.1', '4.4', '4.5', '4.6', '4.7', '4.8', '4.9', '5.1', '5.2', '5.3', '5.4', '6.1', '6.2', '6.3', '6.4', '7.1', '7.2']
```

also, why your `master` branch is ahead of `develop`?

parsifal-47:
sorry, false alarm, I'm using too old version, but I'm still curious about why `master` is ahead of `develop`?

PR #56, opened 2018-01-04
Support C-only projects

DavidZemon:
I was not able to use this in a C-only project due to the reliance on `CMAKE_CXX_COMPILER_VERSION` variable. I've only tested this on Ubuntu 17.10, so I don't know if it will work correctly on MSVC or Apple systems, but I did my best to take educated guesses.

Also, sorry about this going into master rather than develop. I'm not running the v1.0.0 conan release, so I was getting errors while based off of the develop branch because of the missing `-f` flag.


memsharded:
Hi David

You made a great job here. I love to see it working for pure C projects too.

CI is green in all systems:
- https://ci.appveyor.com/project/ConanCIintegration/cmake-conan
- https://travis-ci.org/conan-io/cmake-conan

I don't fully get what was the issue with a ``-f`` flag in 0.30.0, but I am fine merging to master too, by now, people relying on this should be using a tagged release, and this doesn't break them anyway.

Many thanks for this contribution!

Issue #73, opened 2018-02-12
Add Conan recipe as CMake dependency to force reconfigure


DavidZemon:
CMake allows us to force a reconfigure run every time the Conan recipe changes. I think this should be enabled (and I'll submit a pull request in just a moment for this).

The variable to use is `CMAKE_CONFIGURE_DEPENDS`: https://cmake.org/cmake/help/v3.0/prop_dir/CMAKE_CONFIGURE_DEPENDS.html

DavidZemon:
I'm an idiot. Apparently this has worked all along!

PR #74, opened 2018-02-12
[#73] Add Conan recipe as CMake dependency

DavidZemon:
This allows CMake to autodetect when the recipe has changed and rerun the configure step (therefore pulling new dependencies).

### DIFF
--- a/conan.cmake
+++ b/conan.cmake
@@ -178,7 +178,7 @@ function(conan_cmake_settings result)
         conan_cmake_detect_vs_runtime(_vs_runtime)
         message(STATUS "Detected VS runtime: ${_vs_runtime}")
         set(_SETTINGS ${_SETTINGS} -s compiler.runtime=${_vs_runtime})
-        
+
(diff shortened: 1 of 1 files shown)


sixten-hilborn:
This is already handled in `conan_cmake_setup_conanfile` using `configure_file`, does it not work for you?

DavidZemon:
Well whatdya know!? It does work! I can't believe I missed that. I must have just assumed it didn't when I first discovered this repo and never actually _tried_ it. I would personally go with the `set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS "${ARGUMENTS_CONANFILE}")` route over the `configure_file()` route, but I don't think it's deserving of a pull request since it doesn't functionally change anything.

memsharded:
Hi!

Thanks @DavidZemon for the contribution, but as @sixten-hilborn commented, it seems this functionality is already implemented. Should this PR be closed then?

PR #76, opened 2018-02-28
Update README.md to document PROFILE argument

matthew-d-jones:
The `PROFILE` argument is important but not currently mentioned in the README.


memsharded:
The tests were broken, but clearly not your fault. Many thanks for adding this!

PR #86, opened 2018-06-18
Added INSTALL_FOLDER option to conan_install

chuet-thermo:
Hi,

I added the INSTALL_FODLER option of conan_install to conan.cmake, as we require it for the import to copy to the correct folder.

Best


memsharded:
Great, thank you!

Issue #85, opened 2018-06-18
Profile argument causes error.


bc-lee:
If PROFILE or related arguments are passed to `conan_cmake_run` or `conan_cmake_install`, cmake-conan ignores all settings, including `-g cmake`, which makes this option unuseable with CMake integration.

https://github.com/conan-io/cmake-conan/blob/36f25ff678cc32dd95caafe0acde9de34c0f1591/conan.cmake#L313

Following CMakeLists.txt is used with cmake-conan v0.11:
```cmake
cmake_minimum_required(VERSION 3.4)
project(test_cmake_conan)

set(BUILD_CONAN_MISSING FALSE CACHE BOOL "Build missing packages")
set(BUILD_CONAN_PROFILE "" CACHE STRING "Use profile to overide the profile detection")
set(BUILD_CONAN_UPDATE FALSE CACHE BOOL "Check updates exist from upstream remote")

list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/third_party/cmake")

include(conan)
if(BUILD_CONAN_MISSING)
    set(BUILD_CONAN_RULE "missing")
else()
    set(BUILD_CONAN_RULE "")
endif()

if(BUILD_CONAN_UPDATE)
    set(_BUILD_CONAN_UPDATE "UPDATE")
endif()

if(BUILD_CONAN_PROFILE)
    set(CONAN_DISABLE_CHECK_COMPILER True)
    set(_BUILD_CONAN_PROFILE PROFILE "${BUILD_CONAN_PROFILE}")
endif()

string(FIND "${CMAKE_BUILD_TYPE}" "Rel" REL_POS)
if(NOT(REL_POS LESS 0))
    set(_BUILD_CONAN_TYPE "Release")
else()
    set(_BUILD_CONAN_TYPE "Debug")
endif()

conan_cmake_run(
    CONANFILE conanfile.py
    BASIC_SETUP ${_BUILD_CONAN_UPDATE} CMAKE_TARGETS KEEP_RPATHS NO_OUTPUT_DIRS
    BUILD_TYPE ${_BUILD_CONAN_TYPE} ${_BUILD_CONAN_PROFILE}
    BUILD "${BUILD_CONAN_RULE}"
)
```

`cmake /path/to/project -DBUILD_CONAN_PROFILE=default` would print following errors:
```text
$ cmake /path/to/project -DBUILD_CONAN_PROFILE='default'
[...]

PR #87, opened 2018-06-19
 Fix conan_cmake_install to handle PROFILE argument properly 

bc-lee:
Resolves: #85


conan.cmake
```diff
@@ -298,23 +298,44 @@ function(conan_cmake_install)
...
     endif()
     if(ARGUMENTS_PROFILE)
-      set(settings -pr ${ARGUMENTS_PROFILE})
+      set(settings_override -pr ${ARGUMENTS_PROFILE})
     endif()
     if(ARGUMENTS_UPDATE)
       set(CONAN_INSTALL_UPDATE --update)
     endif()
+
+    if(settings_override)
```
memsharded: I am not sure if this is really necessary. The settings naturaly override the profile settings, it is not enough just to concat the strings? What is this code doing?

bc-lee: Okay, I need to explain what I expect conan-cmake would work if `PROFILE` argument is passed. to `conan_cmake_run`.
- `-g cmake` or `-g cmake_multi` should be handled
- At least ` -s build_type` shoud be handled. This is consistent with `conan install` command.

Also, `-s compiler` or `-s compiler.version` should not be handled if `PROFILE` argument is passed. This is different behavior with `conan install` command.
I expect users want to override compiler options, by settings `PROFILE` argument. For example, One can use Clang 5.0/libstdc++11 in Ubuntu 16.04 instead of gcc 5/libstdc++11, but they are binary compatible, so they can pass `PROFILE` for gcc 5/libstdc++11 to use prebuilt libraries.

memsharded: Thanks for the explanation. Still I have concerns to treat differently different settings. The key is that in cmake-conan there are different sources of settings values. I think that the priority should be (from lower to higher):

- Settings auto-detected by the cmake-conan script
- Profile. If a profile is defined, values used in the profile should always have priority to the auto-detected values
- Explicit settings in the cmake-conan script. These should have priority over the profile, in the same way as the command line settings have priority over the profile.

Maybe the approach could be to use ``include(profile)`` feature or something related, but as-is it seems a bit inconsistent right now.

bc-lee: I fully agree with your opinion about the priority for settings. The problem is, `conan_cmake_settings` returns `result` (out) parameter, where some of them are auto-detected settings(like `-s compiler`) and some of them are not. So, in `conan_cmake_install`, it is not possible to determine based on prorities of settings based on its sources.
Maybe one need to refactor some of these stages to achive the source-based priority rules..

memsharded:
Many thanks.

Please follow up there in #104 and provide feedback wrt the syntax, e.g. ``PROFILE_AUTO SETTINGS_COMPILER_VERSION`` vs ``PROFILE_AUTO settings.compiler.version``. Will wait a bit for feedback and do a release soon.

PR #97, opened 2018-08-10
Added an option to quiet conan's output

chuet-thermo:


memsharded:
It seems this was not working as expected, I have proposed a fix for 0.14 using the same OUTPUT_QUIET variable. Please do check if possible when 0.14 is released

PR #98, opened 2018-08-14
Merge profile and settings so settings can override profile.

dewhisna:
This change merges the profile with settings, rather than completely overriding settings.
We had a special case here when using a profile where we needed to also pass a `cppstd` setting to Conan and override/append the profile.  This was our solution.
I think this will solve #85 too.


dewhisna:
Hmmm.. That's weird.  Don't know why it's failing the Unix build on Travis. The tests ran just fine here locally for me.  I am using a different gcc (5.4.0 instead of 4.8.4) and possibly a different cmake (3.11.4).  Looks like there was a transient download error during the run:
```
ERROR: Error downloading file https://api.bintray.com/conan/conan/conan-transit/v1/files/memsharded/Hello/0.1/testing/export/conanfile.py: '('Connection aborted.', BadStatusLine("''",))'. [Remote: transit]
````
Maybe it just needs to be kicked off again?

memsharded:
Hi!

Thanks for submitting this contribution!. I think this collides with other existing PR: https://github.com/conan-io/cmake-conan/pull/87. I am not fully sure of either approach, I am thinking about both and will try to decide or come up with a solution.

memsharded:
This PR has been superseded by https://github.com/conan-io/cmake-conan/pull/104. 
Thanks very much for it, it really helped to spot and throw light to this issue.

PR #104, opened 2018-09-07
profile auto

### LINKED ISSUE #85: Profile argument causes error.

### LINKED ISSUE #96: if BUILD_TYPE is specified, cmake generator should be used instead of cmake_multi
If the user has overridden the default `BUILD_TYPE`, script currently uses `cmake_multi` generator and two times installs the requested `build_type`, which then generates `cmake` configuration error because another `build_type` is missing (i.e. if the user has selected `Debug` build type, the user wants to use it in all configurations within its IDE and CMake throws error that `cmakebuildinfo_release.cmake` is missing).

[Here](https://github.com/microblink/conan-build-helper/blob/master/mb_conan_build.cmake#L136) is my dirty workaround for that issue from wrapper script that invokes your and performs additional setup specific for our internal build system.

### LINKED ISSUE #111: use major-only for Clang >= 8.0
Similar to https://github.com/conan-io/conan/pull/3643.
Instead of passing "compiler.version=${MAJOR}.${MINOR}" to the install command, "compiler.version=${MAJOR}" should be used if Clang >= 8.

memsharded:
This PR implements merging of info between the automatically deduce information from CMake and profiles:

- https://github.com/conan-io/cmake-conan/pull/98
- https://github.com/conan-io/cmake-conan/pull/87

The logic is to define argument PROFILE_AUTO=ARCH, COMPILER, BUILD_TYPE, COMPILER_VERSION.... These are the arguments that will be defined by cmake, overriding them if existing in profiles. It is necessary to introduce it to model the different use cases. Different users want different things: @dewhisna wants that profile and detected settings are always merged, while @bc-lee wants the profile override the detected settings, except for ``arch`` and ``build_type``.

Also close #111 
Also fix #96 
Also close #85


puetzk:
For context I'm a co-worker of @dewhisna. This feels a little wordy, but it **is** nicely explicit. I think I like it.

I would have kept the setting keywords spelled exactly as in conan, though, install of uppercasing them; that would read more like the usual Cmake function(KEYWORD value1 value2) syntax, and cause fewer potential collisions between the keyword args and settings.yml keys (for example, you already have ARCH as a oneValueArgs, so conan_cmake_install (PROFILE_AUTO ARCH) won't actual parse they way I think you intend it (I think it will give an empty ARGUMENTS_PROFILE_AUTO, and an empty ARGUMENTS_ARCH)

I might also suggest that, if you did specify PROFILE_AUTO, it should no longer give the warning 
https://github.com/conan-io/cmake-conan/blob/61ae3b3356f2f2c6ca2a3dd0fedcc21a407bbe1c/conan.cmake#L70-L71

because it will have detected what you asked for, and you didn't want it to be complete anyway. Though I suppose for future-compatibility it should give a warning for unknown setting names (i.e. when _CONAN_SETTING_${ARG} is undefined rather than just falsey) when it loops through ARGUMENTS_PROFILE_AUTO, since that means you asked for something this version doesn't know how to detect.

https://github.com/conan-io/cmake-conan/blob/61ae3b3356f2f2c6ca2a3dd0fedcc21a407bbe1c/conan.cmake#L220-L221



at least if , .

memsharded:
> I would have kept the setting keywords spelled exactly as in conan, though, install of uppercasing them; that would read more like the usual Cmake function(KEYWORD value1 value2) 

Yes, I think this is good, I will try to change it.

About the name collisions, I was wrong in the above example, but right in the code. CMake is case insensitive, so it doesn't matter to say ARCH or "arch", and it was colliding, so I had to define it as ``PROFILE_AUTO SETTINGS_ARCH``, ``PROFILE_AUTO SETTINGS_BUILD_TYPE``. I will try to change it to conan syntax: ``PROFILE_AUTO settings.arch`` and ``PROFILE_AUTO settings.compiler.version``, what do you think?

> I might also suggest that, if you did specify PROFILE_AUTO, it should no longer give the warning

I think maybe it is time to completely remove the warning. It is from earlier versions, where the detection was very un-mature. The detection of the settings is the main feature, and applies to the normal operation without specifying profiles, which is the most popular one for this conan.cmake project. Now it seems it is working reasonably well, so lets remove it.

Thanks for the feeback!

bc-lee:
Using dots in the option (`dotted.lower.case`) is consistent with the conan settings syntax IMO, but it is a matter of preferences. Every other thing looks good to me.

puetzk:
I don't have any problem with settings.arch, but cmake_parse_arguments **is** case-sensitive about its keywords:

```cmake
include (CMakeParseArguments)

cmake_parse_arguments(PREFIX1 "OPTION" "ONE_VALUE" "MULTI_VALUES" MULTI_VALUES OPTION)
cmake_parse_arguments(PREFIX2 "OPTION" "ONE_VALUE" "MULTI_VALUES" MULTI_VALUES option)

message("PREFIX1_OPTION=${PREFIX1_OPTION}")
message("PREFIX1_MULTI_VALUES=${PREFIX1_MULTI_VALUES}")

message("PREFIX2_OPTION=${PREFIX2_OPTION}")
message("PREFIX2_MULTI_VALUES=${PREFIX2_MULTI_VALUES}")
```

prints

```
PREFIX1_OPTION=TRUE
PREFIX1_MULTI_VALUES=
PREFIX2_OPTION=FALSE
PREFIX2_MULTI_VALUES=option
```

Tested in 2.8.12 and 3.12

memsharded:
Excellent @puetzk 
I didn't know cmake was case-sensitive regarding arguments. So I removed ``settings.`` prefix, it is much cleaner. Thanks for the feedback!

puetzk:
It might make sense for PROFILE_AUTO to also accept something like `PROFILE_AUTO "*"` or `PROFILE_AUTO all` or something of that sort. This would make it easier to get the behavior #98 asked for (start with the named profile as default, and then override it with as much detection as possible), though still explicitly asking for it to distinguish this from #87's request. It feels like there ought to be some way to get the same behavior as when using using `[general]default_profile` or `$ENV{CONAN_DEFAULT_PROFILE_PATH}`, just with the cmake caller providing the starting profile name.

Implementing this would just be expanding "*" or "all" (or whatever keyword you want to use) into the same list you have for when there's no PROFILE keyword given, i.e.

https://github.com/conan-io/cmake-conan/blob/6320eb20eb0543b1724cd0cfe05980f774f6c3eb/conan.cmake#L199-L201

PR #113, opened 2018-11-20
0.13

memsharded:


conan.cmake
```diff
@@ -327,6 +327,10 @@ function(conan_cmake_install)
     string (REPLACE ";" " " _conan_args "${conan_args}")
     message(STATUS "Conan executing: ${conan_command} ${_conan_args}")
 
+    if(ARGUMENTS_OUTPUT_QUIET)
+      set(OUTPUT_CONTROL OUTPUT_QUIET)
```
memsharded: HI @chuet-thermo 

I am about to release 0.13, was writing the changelog. Realized that I am not sure how this will inhibit the output. Is this ``OUTPUT_CONTROL`` a special CMake variable? I cannot find it.

Also, the ``execute_process`` has been explicitly changed to output to a variable, that might be used to effectively quiet it.

Could you please have a look, update from "develop" branch and check it? Thanks!

chuet-thermo: Hi,

it seems I made a mistake in my PR, the `OUTPUT_CONTROL` variable was meant to be used as `execute_process`'s `OUTPUT_VARIABLE`.

Based on `develop` it should look like:
```
if(NOT ARGUMENTS_OUTPUT_QUIET)
  message(STATUS "${conan_output}")
endif()
```

Should I make a PR that fixes this ?

memsharded: Yes, please. 0.13 has already been released, but to make sure it is included in next 0.14 release. Many thanks!

PR #124, opened 2019-01-26
Add DOWNLOAD_ONCE sub-command for CMake file()

sergey-bon:
Add reusable `file(DOWNLOAD_ONCE)` extension that works the same way `file(DOWNLOAD)` does, except it doesn't overwrite already existing files. Change `conan.cmake` download scenario from conditional `file(DOWNLOAD)` to `file(DOWNLOAD_ONCE)`.

Also uncommented the download of conan.cmake in CMakeLists.txt example. Is there a reason why local `conan.cmake` used?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,10 @@ project(conan_wrapper CXX)
 
 macro(file action)
     if(${action} STREQUAL DOWNLOAD_ONCE)
+        if(${ARGC} LESS 3)
+            # Too few arguments: show the standard error message
+            _file(DOWNLOAD ${ARGN})
+        endif ()
         if(NOT EXISTS ${ARGV2})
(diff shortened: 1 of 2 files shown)


memsharded:
Hi @sergey-bon 

Thanks for your contribution.

However, I think your code brings more lines and complexity to the script, and the value is not much different for users. The simple ``if(NOT EXIST)`` reads very easy for most users and is more compact.

> Also uncommented the download of conan.cmake in CMakeLists.txt example. Is there a reason why local conan.cmake used?

Yes, there are many users that don't want to download the file from the internet on every new build, so they just download the conan.cmake and put it there.

sergey-bon:
I agree, it is much easier to just copy the `conan.cmake` into your source tree.

PR #127, opened 2019-02-20
Pass the correct build type string to conan

JoshCollins746:
conan install require a case sensitive value for the build_type arguments. CMake is not so fussy about the case used in CMAKE_BUILD_TYPE (DEBUG == Debug == debug).

I am using VisualGDB which sets CMAKE_BUILD_TYPE to DEBUG which causes conan install to fail due to an invalid value for the build_type argument.

This patch will map any case insensitive CMAKE_BUILD_TYPE to the camel case value expected by conan install when creating the settings variable.


memsharded:
Makes sense, thanks for the contribution, and sorry for the delay in handling it.

Issue #131, opened 2019-02-27
conan_check with incorrect output on version mismatch


nwoetzel:
https://github.com/conan-io/cmake-conan/blob/2d4bcdf3672d6d141d3a2401976c264b487c6233/conan.cmake#L509
I just wanted to replace my custom conan installation and version check with the macro "conan_check", but when I use it (with conan 1.12.3 installed)
`conan_check(VERSION 1.12.8 REQUIRED)`
The following is emitted:

> Conan outdated.  Installed: 1.12.8, required: .  Consider updating via 'pip
  install conan --upgrade'.
I would suggest the following modification to the conan_check_macro
````
        if(${CMAKE_MATCH_1} VERSION_LESS ${CONAN_VERSION})
            message(FATAL_ERROR "Conan outdated. Installed: ${CMAKE_MATCH_1}, \
                required: ${CONAN_VERSION}. Consider updating via 'pip \
                install conan==${CONAN_VERSION}'.")
        endif()

memsharded:
Closed by https://github.com/conan-io/cmake-conan/pull/132, will be released in 0.14

PR #132, opened 2019-02-27
correct output for version mismatch in conan_check #131

nwoetzel:
This is the suggest change in issue #131 as pull request


memsharded:
I was also bitten recently by this, thanks very much for fixing this!

PR #135, opened 2019-03-13
Made sure the junk file doesnt suddenly go out of tree if its source

KristianJerpetjon:
If the conanfile is in ../somefolder/conanfile.txt  without this fix it will result in a junk file in ${CMAKE_CURRENT_BINARY_DIR}/../somefolder/conanfile.txt.junk which means it's creating folders outside your build folder


KristianJerpetjon:
Added NO_LOAD to make it possible to opt out from loading conanbuildinfo.cmake
Fixed INSTALL_FOLDER to work when conanbuildinfo.cmake is loaded

memsharded:
Many thanks.
The fixes make sense, but it is not fully clear why not loading the conanbuildinfo.cmake should be an option. I am going to merge this, but I would like to understand. Could you please add a short explanation in the README, which also serves as documentation of the option? Thanks!

Issue #88, opened 2018-06-27
Conan.cmake fails in conan_load_buildinfo if INSTALL_FOLDER is passed as an argument


L-M-V:
conan_load_buildinfo assumes that the conan file is located in the binary dir:
`if(EXISTS "${CMAKE_CURRENT_BINARY_DIR}/${_CONANBUILDINFO}")`
However when passing install_folder this is no longer the case.

I updated the following in my script:
```diff
@@ -381,7 +381,7 @@
 endfunction()
 
 
-macro(conan_load_buildinfo)
+macro(conan_load_buildinfo installFolder)
     if(CONAN_CMAKE_MULTI)
       set(_CONANBUILDINFO conanbuildinfo_multi.cmake)
     else()
@@ -389,9 +389,9 @@
     endif()
     # Checks for the existence of conanbuildinfo.cmake, and loads it
     # important that it is macro, so variables defined at parent scope
-    if(EXISTS "${CMAKE_CURRENT_BINARY_DIR}/${_CONANBUILDINFO}")
-      message(STATUS "Conan: Loading ${_CONANBUILDINFO}")
-      include(${CMAKE_CURRENT_BINARY_DIR}/${_CONANBUILDINFO})
+    if(EXISTS "${installFolder}/${_CONANBUILDINFO}")
+      message(STATUS "Conan: Loading ${installFolder}/${_CONANBUILDINFO}")
+      include(${installFolder}/${_CONANBUILDINFO})
     else()
       message(FATAL_ERROR "${_CONANBUILDINFO} doesn't exist in ${CMAKE_CURRENT_BINARY_DIR}")
     endif()
@@ -422,7 +421,11 @@
         endif()
     endif()
 
-    conan_load_buildinfo()
+    set(CONAN_INSTALL_FOLDER ".")
+    if(ARGUMENTS_INSTALL_FOLDER)
+      set(CONAN_INSTALL_FOLDER ${ARGUMENTS_INSTALL_FOLDER})
+    endif()
+    conan_load_buildinfo(${CONAN_INSTALL_FOLDER})
 
     if(ARGUMENTS_BASIC_SETUP)
         foreach(_option CMAKE_TARGETS KEEP_RPATHS NO_OUTPUT_DIRS)

```

tofferlicious:
We ran into the same issue now. My approach to fix it was this:
```diff
@@ -429,11 +429,16 @@ macro(conan_load_buildinfo)
     else()
[...]

PR #136, opened 2019-03-21
Fix location of conanbuildinfo.cmake (#88)

tofferlicious:
In case ARGUMENTS_INSTALL_FOLDER is set, the location of conanbuildinfo.cmake will be loaded from it with this commit.

Otherwise the method call will fail with an error as described in https://github.com/conan-io/cmake-conan/issues/88


tofferlicious:
It seems like a server is not reachable from the build environments running the tests:
ERROR: Error 502 downloading file https://api.bintray.com/conan/conan/conan-transit/v1/files/memsharded/Hello/0.1/testing/export/conanmanifest.txt

But the file is downloadable for me...

memsharded:
Hi @tofferlicious 

Sorry for the delay. I think this has also been fixed in https://github.com/conan-io/cmake-conan/pull/135, merged to develop. It will be released soon in 0.14, could you please check to make sure? Thanks

tofferlicious:
Hi @memsharded,

Thanks for getting back on this.
Works like a charme :+1:

memsharded:
Great, thanks for the feedback!

PR #138, opened 2019-04-01
More detection of platform settings

ohanar:
This adds more platform detection to settings detection. Currently this only covers a handful of names for various architectures, but more aliases can easily be added in the future.


memsharded:
Hi @ohanar 

thanks for trying to contribute this!
It looks interesting, and I wouldn't oppose to a better detection of architectures. 
However, the tests are broken, it seems that it is mis-detecting some of them. Could you please have a look at the tests? It shouldn't be difficult to run them locally, it is ``nosetests .``, but please ask if you have any question.

ohanar:
The commit that I added should fix the issues that were being encountered. It seems like the Macos build had some network issues with bintray.

memsharded:
This PR exceeds a bit my knowledge of cmake, but it is looking good, passes the tests, so I am going to merge it for next 0.14 release. Many thanks, please stay tuned to see if the new detection breaks some other users 😅

memsharded:
Hi @ohanar 

I have been testing this on my laptop with cmake 3.7, and it fails :(
It seems your PR relies on more modern versions of CMake, something that unfortunately we cannot assume. I am going to try to fix it, otherwise I will revert it.

memsharded:
Reverted, as I said, I don't have enough knowledge of cmake to fully understand what has changed in CMake later than 3.7 and how to make your PR works in previous cmake. Sorry about that, if you could please try to contribute it with better backwards compatibility, that would be nice.

I'd also suggest to use 2 different sections for HOST and not host, instead of a foreach-loop, which makes it a bit more difficult to read and see what applies to what (several if(HOST) inside the loop).

PR #146, opened 2019-05-15
Feature/check always

rddesmond:
Update to use `conan_check` in before calls to conan, to use existing find_program functionality.  This causes `conan_add_remote` and `conan_cmake_install` to honor `CONAN_CMD`, which can be set as an option to cmake.


vcrom:
Hi @rddesmond, probably you want to make the conan_check you added to have the REQUIERED flag.

rddesmond:
@vcrom Good call, that makes the logging more consistent.

conan.cmake
```diff
@@ -535,20 +534,23 @@ macro(conan_check)
...
     #       URL https://api.bintray.com/conan/bincrafters/public-conan)
-    set(oneValueArgs URL NAME INDEX)
+    set(oneValueArgs URL NAME INDEX COMMAND)
     cmake_parse_arguments(CONAN "" "${oneValueArgs}" "" ${ARGN})
 
     if(DEFINED CONAN_INDEX)
         set(CONAN_INDEX_ARG "-i ${CONAN_INDEX}")
     endif()
-
+    if(CONAN_COMMAND)
```
memsharded: I am not sure I fully understand the flow, and what is being set:
- If ``conan_add_remote()`` defines COMMAND, it will define ``CONAN_CMD`` with its value.
- It will then call ``conan_check()`` macro.
- The ``conan_check()`` calls internally ``find_program(CONAN_CMD conan)``. This will try to find in the path the ``conan`` command, not using the argument provided by the user. Furthermore, it will overwrite the ``CONAN_CMD`` variable value.

It seems the same happens above in ``conan_cmake_install()``. I might be missing some cmake-fu, please tell.

rddesmond: There is some magic that here.  In cmake, there is a global variable store, the variable cache.  `find_program(CONAN_CMD conan)` will search first for `CONAN_CMD` on the cache (i.e. in "CMakeCache.txt").  It would be set here with

- set with the `CACHE` flag: `set(CONAN_CMD /path/to/conan CACHE "")`
- cmake from the command line: `cmake -DCONAN_CMD=/path/to/conan ..` (My main motivation for the PR is for this.)
- a previous call to `find_program`

That said, you did highlight a mistake, thank you.  In add remote or cmake install, `CONAN_CMD` is set in the scope of the function and not on the cache.  I put the check in an `else` for the case you highlighted.

memsharded: Great, thanks for the clarification

PR #155, opened 2019-07-07
Pure C projects had errors in detect_vs_runtime()

cmourglia:
CMAKE_CXX_FLAGS* are empty when creating a project with
project(MY_PROJECT C) as a header. This is causing issues
when trying to string(REPLACE) with an empty string.


memsharded:
Yes, good fix, thanks for contributing it!
Lets merge it, will be in next 0.15 release.

PR #157, opened 2019-07-10
Generate multiconfig files for RelWithDebInfo and MinSizeRel

### LINKED ISSUE #156: RelWithDebInfo and MinSizeRel build types not supported with cmake_multi generator
See comments in https://github.com/bincrafters/community/issues/867

I think the problem should be in this line not taking into account the new build types: https://github.com/conan-io/cmake-conan/blob/c6e5f37b2b0ab52ddc5b9c65e14ad3fd6f6b912f/conan.cmake#L477

However, I tried forcing `CMAKE_BUILD_TYPE=RelWithDebInfo` and it worked (with single configuration)

danimtb:
fixes #156


munoah:
would it be possible to only install the build types that are specified in [CMAKE_CONFIGURATION_TYPES](https://cmake.org/cmake/help/latest/variable/CMAKE_CONFIGURATION_TYPES.html)? 
if you only need Release and Debug you shouldn't be forced to have the library installed 4 times (especially if its something like boost and you need to build it yourself because you changed some options?)

danimtb:
You are right, having the CMAKE_CONFIGURATION_TYPES check would be a good thing.

I am not sure if the `RelWithDebInfo` and `MinSizeRel` are forced to be always installed with this change or if it will be done once you build for that config. Please @memsharded confirm this when possible.

Thank you!

danimtb:
Yes, I have checked and probably this will force to install the 4 build types and will be very annoying as we don't have packages for RelWithDebInfo and MinSizeRel in Conan Center. So maybe it is a good opportunity to use CMAKE_CONFIGURATION_TYPES. WDYT @memsharded?

munoah:
i tested it out aswell, the libraries will get built when configuring cmake, however, there are some libraries that cant build, e.G. lz4.
Whatever the behaviour was before conan 1.16.0, it worked, so maybe there should be an option to get that behaviour back? Something like link to release libraries when using relwithdebinfo?
I'll check how libraries get built with non-multi generators, maybe its just a problem with the .sln files that some packages use to build.

noizex:
How are things with this? I'm also experiencing a problem where in RelWithDebInfo nothing seems to get included into the project (I'd assume Release libs would be fine for this, as I don't want to recompile everything with RelWithDebInfo)

czoido:
Hi, I think that using `CMAKE_CONFIGURATION_TYPES` to iterate over the build types (the default value will be `Debug;Release;MinSizeRel;RelWithDebInfo`) would be ok

https://github.com/czoido/cmake-conan/blob/4e35aa3a536c2877acc3dcaf90e97ba3784ae342/conan.cmake#L479

`foreach(CMAKE_BUILD_TYPE ${CMAKE_CONFIGURATION_TYPES})`

and then you can always specify just the build types you want in the CMakeLists.txt
`set(CMAKE_CONFIGURATION_TYPES "RelWithDebInfo;Release")`
 or in the command line:
`cmake .. -DCMAKE_CONFIGURATION_TYPES="RelWithDebInfo;Release"`

noizex:
@czoido Not sure if I get it properly - wouldn't that require building libraries for these build types? RelWithDebInfo is assumed to use the same libraries as Release, and just use different settings per project. I see when I try to `conan install` as RelWithDebInfo it tries to recompile all packages for that build type and has some problems plus it's something I'd rather avoid.

czoido:
Hi @noizex,
That issue (linking `RelWithDebInfo` with `Release` libraries) is more related to Conan itself that considers `RelWithDebInfo`  and `Release` as two completely different configurations that have nothing to do with each other, maybe what you can do is using [compatible packages ](https://docs.conan.io/en/latest/creating_packages/define_abi_compatibility.html#compatible-packages) so that your packages fallback to `Release` configuration if the `RelWithDebInfo` is not found or forcing your dependencies with `build_type=Release` when compiling in `RelWithDebInfo` configuration using `-s *:build_type=Release`
Hope this helps

czoido:
closing as #194 is merged now

Issue #160, opened 2019-07-19
Issues with configure_file trying to replace some patterns


Morwenn:
I have written a `conanfile.py` which happens to contain some bash code in a string, and it makes the whole wrapper fails because of this line:

```
configure_file(${ARGUMENTS_CONANFILE} ${CMAKE_CURRENT_BINARY_DIR}/${_CONANFILE_NAME}.junk)
```

My bash code embedded in a string contains access to arrays elements such as `${MY_ARRAY[0]}`, which `configure_file` tries to replace. If I understand correctly, `configure_file` was used to make sure that changing the `conanfile.py`/`conanfile.txt` would trigger a CMake reexecution, but in my case it's also trying to replace some patterns. Would it be ok to add the option `COPYONLY` to that call to `configure_file` to make sure that it really only copies the file when needed and doesn't try to replace patterns in my `conanfile.py`?

Morwenn:
Closed by #161.

PR #161, opened 2019-07-22
Make configure_file COPYONLY (issue #160)

Morwenn:


memsharded:
Good catch! Were you affected by some unfortunate variable substitution?
Thanks for contributing this

Morwenn:
@memsharded Yes, my issue was explained in #160 :)

memsharded:
Oh, great, didn't see it, thanks!

If you add in the first comment of this PR:

Close #160 

Close|closes|fix|fixes too, then the issue is automatically closed when the PR is merged :)

Morwenn:
I use that in my own projects but tend not to do it for other people's projects just in case they want to accept the merge request but feel like there's more to do to fix the corresponding issue.

PR #165, opened 2019-07-27
add missing `=` in profile command line argument

tkhyn:
Adding this `=` after `-pr` enables the PROFILE passed to conan_cmake_run to be used. Previously (at least on my Windows test machine), the PROFILE was not taken into account and conan was running with the default profile. Now the profile used is the correct one.

Note: this makes the treatment of the `-pr` argument consistent with other conan.cmake-injected arguments such as `-g` or `--build`, which both have an `=` sign.

### DIFF
--- a/conan.cmake
+++ b/conan.cmake
@@ -220,7 +220,7 @@ function(conan_cmake_settings result)
     endif()
 
     foreach(ARG ${_APPLIED_PROFILES})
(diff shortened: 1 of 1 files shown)


memsharded:
Thanks for fixing this! Will be in next 0.15 release.

PR #172, opened 2019-08-22
support for emscripten

DoDoENT:
- allow emscripten as supported platform
- skip libcxx detection with emscripten, as it will fail with `em++` - simply set `libcxx`, as it's the only possible option at the moment


czoido:
Hi @DoDoENT,
Sorry for the late response and thanks a lot for the contribution.
I think that this should be addressed by these two PR that have already been merged:

- https://github.com/conan-io/cmake-conan/pull/258 Adding `SETTING compiler.libcxx libc++` as an argument should work
- https://github.com/conan-io/cmake-conan/pull/232 `Emscripten` platform was added

I'm closing this PR, but please report if you have any issues.

Issue #182, opened 2019-10-23
Automatic conan settings detection issue


wawa19933:
## Summmary

There is an issue, when using `conan.cmake` in subdirectory, after `conan_basic_setup` is already invoked in parent project.

Below is the project structure:

```
/test-project
├── CMakeLists.txt
├── conanfile.py
└── sub-project
   ├── CMakeLists.txt
   └── conanfile.py
```

<details><summary>See CMake configure log</summary>
<p>

```log
-- The CXX compiler identification is GNU 9.2.0
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- COMPILE_OPTIONS= | COMPILE_DEFINITIONS= | /test-project
-- Downloading conan.cmake [v0.14] from https://github.com/conan-io/cmake-conan
-- Conan: checking conan executable in path
-- Conan: Found program /usr/bin/conan
-- Conan: Version found Conan version 1.18.4

-- Conan: Automatic detection of conan settings from cmake
-- Conan: Settings= -s;build_type=Release;-s;compiler=gcc;-s;compiler.version=9;-s;compiler.libcxx=libstdc++11
-- Conan executing: conan install /test-project/conanfile.py -s build_type=Release -s compiler=gcc -s compiler.version=9 -s compiler.libcxx=libstdc++11 -g=cmake -if=/test-project/cmake-build-test
Configuration:
[settings]
arch=x86_64
arch_build=x86_64
build_type=Release
compiler=gcc
compiler.libcxx=libstdc++11
compiler.version=9
os=Linux
os_build=Linux
[options]
[build_requires]
[env]

conanfile.py (alice/0.1.1): Installing package
Requirements
Packages

conanfile.py (alice/0.1.1): Generator cmake created conanbuildinfo.cmake
conanfile.py (alice/0.1.1): Generator txt created conanbuildinfo.txt
conanfile.py (alice/0.1.1): Generated conaninfo.txt
conanfile.py (alice/0.1.1): Generated graphinfo
-- Conan: Loading conanbuildinfo.cmake
-- Conan: Adjusting output directories
-- Conan: Using cmake global configuration
-- Conan: Adjusting default RPATHs Conan policies
-- Conan: Adjusting language standard
[...]

Issue #140, opened 2019-04-18
Wrong setting for 'settings.compiler.libcxx' detected for gtest/1.8.1@bincrafters/stable


lorenzzhao:
When trying to add dependency to gtest:

```
conan_cmake_run(
    REQUIRES
    gtest/1.8.1@bincrafters/stable
    BASIC_SETUP
    )
```

on call of 

`cmake .. -DCMAKE_BUILD_TYPE=Debug`

I get:

```
-- Conan: Adjusting output directories
-- Conan: Using cmake global configuration
-- Conan: Adjusting default RPATHs Conan policies
-- Conan: Adjusting language standard
-- Current conanbuildinfo.cmake directory: /home/.../build-test-Debug
-- Conan: Compiler GCC>=5, checking major version 5
-- Conan: Checking correct version: 5
-- Conan: C++ stdlib: libstdc++
-- Conan: Automatic detection of conan settings from cmake
c++: error: $<$<CONFIG:Debug>:>: No such file or directory
c++: error: $<$<CONFIG:Release>:>: No such file or directory
c++: error: $<$<CONFIG:RelWithDebInfo>:>: No such file or directory
c++: error: $<$<CONFIG:MinSizeRel>:>: No such file or directory
-- Conan: Settings= -s;build_type=Debug;-s;compiler=gcc;-s;compiler.version=5;-s;compiler.libcxx=libc++
-- Conan executing: conan install . -s build_type=Debug -s compiler=gcc -s compiler.version=5 -s compiler.libcxx=libc++ -g=cmake
-- ERROR: Invalid setting 'libc++' is not a valid 'settings.compiler.libcxx' value.
Possible values are ['libstdc++', 'libstdc++11']
Read "http://docs.conan.io/en/latest/faq/troubleshooting.html#error-invalid-setting"

CMake Error at /home/.../cmake/conan.cmake:385 (message):
  Conan install failed='1'
Call Stack (most recent call first):
  /home/.../cmake/conan.cmake:465 (conan_cmake_install)
  /home/.../CMakeLists.txt:12 (conan_cmake_run)

```

Working on Ubuntu 16.04 LTS with default GCC (5.4.0).

Croydon:
Please have a look in your `settings.yml` file. It looks like only `['libstdc++', 'libstdc++11']` would be the possible values in there for `settings.compiler.libcxx`

You need to add `libc++` if you want to use it

memsharded:
[...]

PR #183, opened 2019-10-23
Fix compile_options before run compiler

wawa19933:
I propose a fix for #182 

It will filter out all generator expressions before invoking compiler for automatic checking C++ STD library on UNIX

### DIFF
--- a/conan.cmake
+++ b/conan.cmake
@@ -263,6 +263,7 @@ function(conan_cmake_detect_unix_libcxx result)
             endif()
         endif()
     endforeach()
+    string(GENEX_STRIP "${compile_options}" compile_options)
 
     execute_process(
         COMMAND ${CMAKE_COMMAND} -E echo "#include <string>"


verbatom:
What do you think about using [daedric suggestion](https://github.com/conan-io/cmake-conan/issues/140#issuecomment-526100306) of using the dedicated method `string(GENEX_STRIP "${compile_options} compile_options)` instead of hand-crafted regular expressions?

wawa19933:
> What do you think about [daedric's suggestion](https://github.com/conan-io/cmake-conan/issues/140#issuecomment-526100306) of using the dedicated method `string(GENEX_STRIP "${compile_options}" compile_options)` instead of hand-crafted regular expressions?

Completely agree, this way is much better. `GENEX_STRIP` has also better backward compatibility. 

And thank you, for referencing duplicate issue.

puetzk:
I would like to propose that  (my) PR #197 replace this one for milestone 0.16; it includes the `STRIP_GENEX` for `COMPILE_OPTIONS` found here, but also fixes a similar issue with `COMPILE_DEFINITIONS`, and also missing options for cross-compilers that use `CMAKE_CXX_COMPILER_ARG1` or `CMAKE_CXX_COMPILE_OPTIONS_SYSROOT`. They'll conflict if anyone tries to merge both.

wawa19933:
I can confirm it works for the same issue #182 

> I would like to propose that (my) PR #197 replace this one for milestone 0.16; it includes the `STRIP_GENEX` for `COMPILE_OPTIONS` found here, but also fixes a similar issue with `COMPILE_DEFINITIONS`, and also missing options for cross-compilers that use `CMAKE_CXX_COMPILER_ARG1` or `CMAKE_CXX_COMPILE_OPTIONS_SYSROOT`. They'll conflict if anyone tries to merge both.

czoido:
Closing this as https://github.com/conan-io/cmake-conan/pull/197 has been merged.

PR #184, opened 2019-11-08
Fix work with conan pip install to home folder

antlad:


conan.cmake
```diff
@@ -516,7 +516,7 @@ macro(conan_check)
     set(oneValueArgs VERSION)
     cmake_parse_arguments(CONAN "${options}" "${oneValueArgs}" "" ${ARGN})
 
-    find_program(CONAN_CMD conan)
```
sergeyklay: AFAIK it's work w/o this change. Just add $HOME/.local/bin to the user's PATH. Isn't?

antlad: Nope, as I remember wasn't working. Because conan itself will not work without adding $HOME/.local/bin to PATH when it is installed in home folder. So this was needed fix for cmake.

czoido:
Hi @antlad, you can pass directly the path to Conan in `conan_cmake_run`, wouldn't it solve your problem?

```
conan_cmake_run(REQUIRES fmt/6.1.2
                BASIC_SETUP CMAKE_TARGETS
                BUILD missing
                CONAN_COMMAND "path_to_conan")

```

czoido:
I'm closing this as `CONAN_COMMAND` argument should work to use conan from any location. It has been documented in https://github.com/conan-io/cmake-conan/pull/264
Please feel free to open another issue if your problems still persist.

PR #191, opened 2019-12-23
Auto detect android platform settings

akemimadoka:


akemimadoka:
We can also detect NDK path by examine ANDROID_NDK variable, which is set in toolchain file in NDK, and set
 CONAN_CMAKE_ANDROID_NDK, should it be set here?

PR #192, opened 2019-12-28
Honour the SKIP_STD argument

smessmer:


czoido:
Hi @smessmer,
You are right the `SKIP_STD` was not being considered, we will add this for the next release.
Thanks a lot

PR #199, opened 2020-02-07
Finding conanbuildinfo in either CMAKE_BINARY_DIR or CMAKE_CURRENT_BINARY_DIR

bernedom:
If a project is using cmake-conan in a CMakeLists.txt that is not the root-CMakeLists included using `add_subdirectory(...)` this might cause the build to fail on some plattforms (testted windows using msvc).  
The underlying problem is that a pure cmake build will put the `conanbuildinfo.cmake` into the relative path in the build folder while if the build is invoked using `conan create` the buildinfo file is in the root of the build folder. 

The problem might be with conan itself, but having multiple possible options to look for will fix the problem as well. 

See [SI](https://github.com/bernedom/SI) for an example of the setup I'm referring to.

### DIFF
--- a/conan.cmake
+++ b/conan.cmake
@@ -453,7 +453,10 @@ macro(conan_load_buildinfo)
     if(ARGUMENTS_INSTALL_FOLDER)
         set(_CONANBUILDINFOFOLDER ${ARGUMENTS_INSTALL_FOLDER})
     else()
-        set(_CONANBUILDINFOFOLDER ${CMAKE_CURRENT_BINARY_DIR})
+        find_path(_CONANBUILDINFOFOLDER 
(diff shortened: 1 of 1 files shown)


czoido:
Hi @bernedom, 
Thanks a lot for the contribution and sorry for the late response.
Have you tried to use [CONAN_EXPORTED](https://github.com/conan-io/cmake-conan#creating-packages) to get the conanbuildinfo from the folder you want if you use a `conan create` ?
I think that maybe that could solve your issue. Please report if that fits your use case. Thanks!

bernedom:
Hi @czoido 
Thanks for your answer, that is the hint I needed. 
Working with `CONAN_EXPORTED` solves the problem. That makes this PR obsolete.

PR #215, opened 2020-03-23
fixed argument parsing in conan_config_install

juanma-brl:
I had to modify the macro conan_config_install to be able to parse arguments properly. Otherwise conan config install will not work as soon as there are more parameters than ITEM.


czoido:
Thanks a lot for the contribution @juanma-brl 
This has been added for 0.16 release. Sorry for being so late to review.

czoido (approved):

memsharded (approved):

Issue #216, opened 2020-03-25
Example in the Readme broken?


mristin:
Hi,
I copy/pasted the example you provided in the readme and tried to compile. Unfortunately, the compilation failed. It seems that `Hello/0.1@memsharded/testing` is not available any more.

This is the example CMakeLists.txt I used:
```cmake
cmake_minimum_required(VERSION 2.8)
project(myproject CXX)

# Download automatically, you can also just copy the conan.cmake file
if(NOT EXISTS "${CMAKE_BINARY_DIR}/conan.cmake")
   message(STATUS "Downloading conan.cmake from https://github.com/conan-io/cmake-conan")
   file(DOWNLOAD "https://github.com/conan-io/cmake-conan/raw/v0.15/conan.cmake"
                 "${CMAKE_BINARY_DIR}/conan.cmake")
endif()

include(${CMAKE_BINARY_DIR}/conan.cmake)

conan_cmake_run(REQUIRES Hello/0.1@memsharded/testing
                BASIC_SETUP
                BUILD missing)

add_executable(main main.cpp)
target_link_libraries(main ${CONAN_LIBS})
```

I took a hello-world program for `main.cpp`.

And this is the error:

```
$ cmake .. -DCMAKE_BUILD_TYPE=Release
-- Conan: Automatic detection of conan settings from cmake
-- Conan: Settings= -s;build_type=Release;-s;compiler=gcc;-s;compiler.version=7;-s;compiler.libcxx=libstdc++11
-- Conan: checking conan executable
-- Conan: Found program /home/marko/workspace/zhaw-ims/scapremai-preliminary/cpp-dev-venv/bin/conan
-- Conan: Version found Conan version 1.23.0

-- Conan executing: /home/marko/workspace/zhaw-ims/scapremai-preliminary/cpp-dev-venv/bin/conan install . -s build_type=Release -s compiler=gcc -s compiler.version=7 -s compiler.libcxx=libstdc++11 -g=cmake --build=missing
Configuration:
[settings]
arch=x86_64
arch_build=x86_64
build_type=Release
compiler=gcc
compiler.libcxx=libstdc++11
compiler.version=7
os=Linux
os_build=Linux
[options]
[build_requires]
[env]

Hello/0.1@memsharded/testing: Not found in local cache, looking in remotes...
Hello/0.1@memsharded/testing: Trying with 'conan-center'...
ERROR: Unable to find 'Hello/0.1@memsharded/testing' in remotes
[...]

PR #217, opened 2020-03-25
Fix example

czoido:
Closes: #216


CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
-cmake_minimum_required(VERSION 2.8)
-project(conan_wrapper CXX)
+cmake_minimum_required(VERSION 2.8.12)
+project(MD5Encrypter)
```
danimtb: Maybe change the name to ``MD5hasher`` or something?

Issue #159, opened 2019-07-19
Detects AppleClang as Clang


mathiasi:
Seems like this has been discussed before: https://github.com/conan-io/cmake-conan/issues/52 , https://github.com/conan-io/cmake-conan/issues/33

Even though they have been marked as resolved, it's still happening to me right now on version 0.14.
Let me know if you need other information and I will happily provide to resolve this issue! :)

```
$ cmake .. -DCMAKE_BUILD_TYPE=Release
-- The C compiler identification is AppleClang 10.0.0.10001145
-- The CXX compiler identification is AppleClang 10.0.0.10001145
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Downloading conan.cmake from https://github.com/conan-io/cmake-conan
-- Conan: Automatic detection of conan settings from cmake
In file included from <stdin>:1:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/string:477:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/string_view:176:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/__string:56:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/algorithm:641:
[...]

PR #218, opened 2020-03-26
silence macos warning

timblechmann:
i was running into this compiler invocation error on two different macos computers. the error seems harmless, but nevertheless it's ugly and confuses devs.

i don't exactly like the fix of adding an implicit preprocessor path, but otoh the change is not very intrusive. thoughts?

----

on macos we are running into the following warning when trying to run
the system autodetection:

```
/Applications/Xcode11.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/string.h:61:15:
fatal error: 'string.h' file not
      found
              ^~~~~~~~~~
```

it seems that the compiler cannot find the sdk root, so the c++ headers
cannot find the c headers. we therefore add the macos sdk path manually

### DIFF
--- a/conan.cmake
+++ b/conan.cmake
@@ -264,6 +264,13 @@ function(conan_cmake_detect_unix_libcxx result)
         endif()
     endforeach()
 
+    if (APPLE)
+        # ensure that <string> can include <string.h>
+        if (IS_DIRECTORY /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include)
+            set(compile_options "${compile_option}" "-I/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/")
+        endif()
+    endif()
+
     execute_process(
         COMMAND ${CMAKE_COMMAND} -E echo "#include <string>"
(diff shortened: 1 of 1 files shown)


czoido:
Hi @timblechmann,
Sorry for the late response.
Have you tried to set: `export SDKROOT=$(xcodebuild -version -sdk macosx Path)` (See: https://github.com/conan-io/cmake-conan/issues/159#issuecomment-519486541)
Please check if that solves your issue.

czoido:
Related to: https://github.com/conan-io/cmake-conan/issues/159

timblechmann:
@czoido observations:
*  the warning does not seem to trigger with xcode-11.6 (i had been using this patch locally for months). it does trigger for xcode-10 though (and in the past it was triggered for 11.2)
* running the echo | clang pipeline from terminal neither shows the error on either of these compilers (which is funny)
* setting the SDKROOT environment variable seems to work around it (though i'd appreciate if i wouldn't need to do it manually)

czoido:
Hi @timblechmann we have added this: https://github.com/conan-io/cmake-conan/pull/226 to the 0.16 milestone and that will also address this error. I'll close this one. Please check if that solves the issue, otherwise please feel free to reopen.
Thanks a lot

Issue #225, opened 2020-04-26
Compile error on configuration on MacOS


dforsten:
On MacOS I am getting these errors on executing `conan_cmake_run`
```
-- Conan: Automatic detection of conan settings from cmake
In file included from <stdin>:1:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/string:504:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/string_view:175:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/__string:56:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/algorithm:641:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/cstring:60:
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/string.h:60:15: fatal error: 'string.h' file not found
#include_next <string.h>
              ^~~~~~~~~~
1 error generated.
```

The errors are caused by this command:
```
    execute_process(
        COMMAND ${CMAKE_COMMAND} -E echo "#include <string>"
        COMMAND ${CMAKE_CXX_COMPILER} -x c++ ${compile_options} -E -dM -
        OUTPUT_VARIABLE string_defines
    )
```

Where `${CMAKE_CXX_COMPILER}` is expanding to `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++`. Unlike for `/usr/bin/c++`  the sysroot is not set when executing that binary, but setting the sysroot manually fixes the compile issue:
```
    execute_process(
        COMMAND ${CMAKE_COMMAND} -E echo "#include <string>"
        COMMAND ${CMAKE_CXX_COMPILER} -x c++ --sysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk ${compile_options} -E -dM -
        OUTPUT_VARIABLE string_defines
    )
```

dforsten:
[...]

PR #226, opened 2020-04-26
Fix for cmake-conan issue #225 - Compile error on configuration on MacOS

dforsten:
The sysroot option is not set when using `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++`, needs to be added explicitly.

Closes: https://github.com/conan-io/cmake-conan/issues/159
Closes: https://github.com/conan-io/cmake-conan/issues/225

### DIFF
--- a/conan.cmake
+++ b/conan.cmake
@@ -264,9 +264,13 @@ function(conan_cmake_detect_unix_libcxx result)
         endif()
     endforeach()
 
+    if(APPLE)
+        set(xcode_sysroot_option "--sysroot=${CMAKE_OSX_SYSROOT}")
+    endif()
+
     execute_process(
         COMMAND ${CMAKE_COMMAND} -E echo "#include <string>"
-        COMMAND ${CMAKE_CXX_COMPILER} -x c++ ${compile_options} -E -dM -
+        COMMAND ${CMAKE_CXX_COMPILER} -x c++ ${xcode_sysroot_option} ${compile_options} -E -dM -
         OUTPUT_VARIABLE string_defines
     )
 
--- a/conan.cmake
+++ b/conan.cmake
@@ -264,7 +264,7 @@ function(conan_cmake_detect_unix_libcxx result)
         endif()
     endforeach()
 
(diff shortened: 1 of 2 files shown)


conan.cmake
```diff
@@ -264,9 +264,13 @@ function(conan_cmake_detect_unix_libcxx result)
         endif()
     endforeach()
 
+    if(APPLE)
+        set(xcode_sysroot_option "--sysroot=${CMAKE_OSX_SYSROOT}")
```
memsharded, line 268: Is it possible that CMAKE_OSX_SYSROOT is not defined? For example in iOS? I would say it is possible, and this would break for some users not in OSX. Does it make sense to test for CMAKE_OSX_SYSROOT existence instead?

(sorry this took a while to review, Conan amazing growth keeps the team very busy)

czoido:
Hi @dforsten,
Sorry for the late response.
Have you tried to set: `export SDKROOT=$(xcodebuild -version -sdk macosx Path)` (See: https://github.com/conan-io/cmake-conan/issues/159#issuecomment-519486541)
Please check if that solves your issue.

czoido:
Related to: https://github.com/conan-io/cmake-conan/issues/159

conan.cmake
czoido: I also would suggest to check `CMAKE_OSX_SYSROOT` instead of `APPLE`.
Also @dforsten I have added a test for this issue.

conan.cmake
```diff
@@ -264,9 +264,13 @@ function(conan_cmake_detect_unix_libcxx result)
         endif()
     endforeach()
 
+    if(APPLE)
```
czoido, line 267:
```suggestion
    if(CMAKE_OSX_SYSROOT)
```

dforsten:
Cool!

Trying to get Conan/Artifactory used more and more here at vizrt,
thanks for writing such a great tool! :-)

PR #228, opened 2020-05-04
Add VERIFY_SSL arg to conan_add_remote

gordonread:
This adds a VERIFY_SSL argument to the conan_add_remote function (We have a local conan repository that requires the verify ssl argument be false)

### DIFF
--- a/conan.cmake
+++ b/conan.cmake
@@ -549,24 +549,28 @@ endmacro()
 
 function(conan_add_remote)
     # Adds a remote
-    # Arguments URL and NAME are required, INDEX and COMMAND are optional
+    # Arguments URL and NAME are required, INDEX, COMMAND and VERIFY_SSL are optional
     # Example usage:
     #    conan_add_remote(NAME bincrafters INDEX 1
-    #       URL https://api.bintray.com/conan/bincrafters/public-conan)
-    set(oneValueArgs URL NAME INDEX COMMAND)
+    #       URL https://api.bintray.com/conan/bincrafters/public-conan
+    #       VERIFY_SSL True)
+    set(oneValueArgs URL NAME INDEX COMMAND VERIFY_SSL)
     cmake_parse_arguments(CONAN "" "${oneValueArgs}" "" ${ARGN})
 
     if(DEFINED CONAN_INDEX)
         set(CONAN_INDEX_ARG "-i ${CONAN_INDEX}")
     endif()
-    if(CONAN_COMMAND)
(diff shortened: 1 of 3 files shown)


conan.cmake
```diff
@@ -549,24 +549,28 @@ endmacro()
...
+        set(CONAN_CMD ${CONAN_COMMAND})
     else()
         conan_check(REQUIRED)
     endif()
-    message(STATUS "Conan: Adding ${CONAN_NAME} remote repository (${CONAN_URL})")
-    execute_process(COMMAND ${CONAN_CMD} remote add ${CONAN_NAME} ${CONAN_URL}
-      ${CONAN_INDEX_ARG} -f)
+    set(CONAN_VERIFY_SSL_ARG "True")
+    if(DEFINED CONAN_VERIFY_SSL)
+        set(CONAN_VERIFY_SSL_ARG ${CONAN_VERIFY_SSL})
```
memsharded, line 570: It seems this is missing the ``--verify-ssl`` argument. In the same way the ``CONAN_INDEX_ARG`` introduces the ``-i ...`` argument, this will need this (note the other args as NAME, URL are positional instead)

memsharded (changes requested):
Hi @gordonread 

It seems there is something wrong in this contribution. Possibly adding a test would have raised this error, could you please try to add such test?

Many thanks! (and sorry this took a while to review)

conan.cmake
gordonread: Thanks for looking at this, for me, conan reports:

```conan remote add
usage: conan remote add [-h] [-i [INSERT]] [-f] remote url [verify_ssl]
conan remote add: error: the following arguments are required: remote, url
ERROR: Exiting with code: 2
```
so this change simply sets `${CONAN_VERIFY_SSL_ARG}` to either True or whatever the `VERIFY_SSL` parameter is set to if it's there, this value is then put on the end of the command (below) as per the help above - I can't see any mention of a `--verify-ssl` in the conan docs that I have.
Please correct me if I'm wrong.
I'll add a test as requested in your email though.
/Gordon

czoido: I think this should be ok, now the `verify_ssl` argument for conan remote add is positional: https://github.com/conan-io/conan/blob/29b04421a650560673db9e41a9d62060bda3dc90/conans/client/command.py#L1509-L1510
@gordonread I'm adding a test to check the changes

czoido (approved):

Issue #229, opened 2020-05-08
compiler.libcxx is not properly recognized from CMake parameters


Scarabytes:
I'm using Clang 8 and libc++ on a Linux Mint 18.3 system. I tell CMake via the CC, CXX and CXXFLAGS environment variables (or the CMAKE_* variables) to use Clang and libc++ instead of GCC and libstdc++. However, conan-cmake still seems to prefer libstdc++ over libc++. Am I doing something wrong, or is this maybe a bug?

My conan code (which is run at the very beginning of the CMake project):
``` cmake
if(NOT EXISTS "${CMAKE_BINARY_DIR}/conan.cmake")
  message(STATUS "Downloading conan.cmake from https://github.com/conan-io/cmake-conan/tree/v0.15")
  file(DOWNLOAD "https://github.com/conan-io/cmake-conan/raw/v0.15/conan.cmake"
                "${CMAKE_BINARY_DIR}/conan.cmake"
                STATUS DOWNLOAD_STATUS)

  list(GET DOWNLOAD_STATUS 0 ERRCODE)
  list(GET DOWNLOAD_STATUS 1 MESSAGE)

  if (NOT ${ERRCODE} EQUAL 0)
    message(FATAL_ERROR "Download failed: ${MESSAGE}")
    file(REMOVE "${CMAKE_BINARY_DIR}/conan.cmake")
  else()
    message(STATUS "Downloaded to ${CMAKE_BINARY_DIR}/conan.cmake")
  endif()
endif()

include(${CMAKE_BINARY_DIR}/conan.cmake)

conan_cmake_run(
  CONANFILE ../conanfile.txt
  BUILD missing
  BASIC_SETUP
  CMAKE_TARGETS
)
```

The output when invoking cmake:
Note that I explicitly set the CXX Flags to contain `-stdlib=libc++`. The compiler does use this standard library. Conan however detects `libstdc++11`.
```
$ cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_COMPILER=clang-8 -DCMAKE_CXX_COMPILER=clang++-8 -DCMAKE_CXX_FLAGS=-stdlib=libc++ ..

-- The CXX compiler identification is Clang 8.0.1
-- Check for working CXX compiler: /usr/bin/clang++-8
-- Check for working CXX compiler: /usr/bin/clang++-8 -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
[...]

PR #197, opened 2020-01-24
Add additional compiler options to conan_cmake_detect_unix_libcxx

puetzk:
Without these many options many wrappers (e.g. ccache, winegcc) or cross-compiling toolchains (e.g. yocto) may not find the intended underlying tools and sysroot, leading to the command failing, or (perhaps worse) succeeding, but using defaults instead of the intended underlying gcc or libstdc++ and thus getting inaccurate results.

Fixes #182
Fixes #195

PR #233, opened 2020-05-18
use stdlib clang flag to check libcxx

czoido:
When trying to detect the `libcxx` setting for conan if using clang you can set the `-stdlib` flag to `libc++` or `libdstdc++` but that was not being taken into account if the flag was set using `CMAKE_CXX_FLAGS` (but it was used if the flag was set with `add_compile_options`) 

Fixes: #229

### DIFF
--- a/conan.cmake
+++ b/conan.cmake
@@ -264,6 +264,14 @@ function(conan_cmake_detect_unix_libcxx result)
         endif()
     endforeach()
 
+    if (${CMAKE_${LANGUAGE}_COMPILER_ID} STREQUAL AppleClang OR ${CMAKE_${LANGUAGE}_COMPILER_ID} STREQUAL Clang)
+        foreach(cxx_flag ${CMAKE_CXX_FLAGS})
+            if(cxx_flag MATCHES "stdlib")
+                set(compile_options ${compile_options} ${cxx_flag})
+            endif()
+        endforeach()
+    endif()
+
     execute_process(
         COMMAND ${CMAKE_COMMAND} -E echo "#include <string>"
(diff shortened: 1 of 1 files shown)


smessmer:
This should probably also set the `libcxx` variable in `PARENT_SCOPE`, see https://github.com/conan-io/cmake-conan/issues/256

puetzk:
CMAKE_CXX_FLAGS is a string variable, not a ;-list. So you can't use foreach() on it directly, you have to deal with the shell quoting. And even that aside, I think the regex ought to be stricter, e.g. `if(cxx_flag MATCHES "(^| )-stdlib=")`? As is this could mismatch in the middle of some unrelated argument if e.g. a filename contained the word stdlib. c.f. how CMake itself derives this in CMakeDetermineCompilerId, though (unfortuntately) it only serializes it for Xcode projects, so you can't just use `${CLANG_CXX_LIBRARY}`.

https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/CMakeDetermineCompilerId.cmake#L508-516
https://gitlab.kitware.com/cmake/cmake/-/commit/8af334f5baa1a628895571491f67a4f7d9483d9b

But, conveniently, I actually think this whole PR is moot now that you accepted #197. It introduced using `separate_arguments` to parse `${CMAKE_CXX_FLAGS}` and passing the whole thing along to the `conan_cmake_detect_unix_libcxx` probing. So if there was a `-stdlib` in there you've already got it.

czoido:
Hi @puetzk,
Thanks a lot for the comment, I think that's probably right, I'll do some tests and close the PR in that case.

czoido:
This PR is no longer needed after the merge of #197.

Issue #238, opened 2020-05-24
conan_config_install() fails with optional argumens


Sirdavos:
`conan_config_install(ITEM /locate/to/file.zip VERIFY_SSL False)`
Once above command is called in cmake, below error is generated. 
usage: conan config [-h] {get,home,install,rm,set} ... 
conan config: error: unrecognized arguments
ERROR: Exiting with code: 2

I found the issue is related with conan.cmake
In macro(conan_config_install) 
.. 
`    if(DEFINED CONAN_VERIFY_SSL)
        set(CONAN_CONFIG_INSTALL_ARGS "${CONAN_CONFIG_INSTALL_ARGS} --verify-ssl=${CONAN_VERIFY_SSL}")
    endif()`

I removed **white space** before --verify-ssl  and the issue is solved.

czoido:
Hi @Sirdavos,
Thanks a lot for reporting the issue, looks like there were some quotes missing when calling to `execute_process`. I'll open a PR to fix the issue.

PR #241, opened 2020-05-26
Fix/238

czoido:
Fixes: #238


conan.cmake
```diff
@@ -586,25 +586,24 @@ macro(conan_config_install)
     endif()
 
     if(DEFINED CONAN_VERIFY_SSL)
-        set(CONAN_CONFIG_INSTALL_ARGS "${CONAN_CONFIG_INSTALL_ARGS} --verify-ssl ${CONAN_VERIFY_SSL}")
+        string(STRIP "${CONAN_CONFIG_INSTALL_ARGS} --verify-ssl=${CONAN_VERIFY_SSL}" CONAN_CONFIG_INSTALL_ARGS)
```
memsharded, line 589: This way of concatenating args, using a STRIP is a bit weird. I think I would have concat (using "=", that is correct), and then strip at the end if necessary. Am I missing something?

memsharded (approved):

conan.cmake
czoido: Yes, it's a bit weird, in fact, I have just realized that there was another PR from an user solving this: https://github.com/conan-io/cmake-conan/pull/215 I'll close this one in favor of that one and add the test there.

PR #244, opened 2020-06-02
no autodetection of settings if profile is specified directly as

petermbauer:
Although a profile is given to the `conan_cmake_run` function it tries to autodetect settings from cmake. In case of `CMAKE_SYSTEM_NAME` being `Generic` this leads to a fatal error "cmake system Generic is not supported by conan ...". I propose to not perform the autodetection at all in case the profile is given as argument.
IMO this is also according to the documentation (https://github.com/conan-io/cmake-conan/blame/develop/README.md#L144-L146).


czoido:
Hi @petermbauer,
Thanks a lot for reporting the issue.
According to this: https://github.com/conan-io/cmake-conan/blob/c29f0021c22bb1b91eeccf3c9b96a23b5cb8524b/conan.cmake#L227
No auto-detection will be made if you define the profile.
The problem is that here: https://github.com/conan-io/cmake-conan/blob/c29f0021c22bb1b91eeccf3c9b96a23b5cb8524b/conan.cmake#L106 the latest supported conan platforms are hardcoded, I would think that you could add your platform to the `settings.yaml` as a custom setting but that will also fail unless you add it to that hardcoded list. We'll have a look at some possible solutions for the issue.

czoido:
Hi @petermbauer,
I'm closing this one as this PR: https://github.com/conan-io/cmake-conan/pull/292 should fix the original problem.

PR #250, opened 2020-06-16
Add build and host profiles

cschlosser:
Add option to make conan detection silent since it's triggered
three times otherwise when build and host profiles are specified.

### DIFF
--- a/conan.cmake
+++ b/conan.cmake
@@ -223,13 +223,9 @@ function(conan_cmake_settings result)
     foreach(ARG ${_APPLIED_PROFILES})
         set(_SETTINGS ${_SETTINGS} -pr=${ARG})
     endforeach()
-    foreach(ARG ${ARGUMENTS_BUILD_PROFILE})
+    foreach(ARG ${ARGUMENTS_PROFILE_BUILD})
         conan_check(VERSION 1.24.0 REQUIRED DETECT_QUIET)
         set(_SETTINGS ${_SETTINGS} -pr:b=${ARG})
-    endforeach()
-    foreach(ARG ${ARGUMENTS_HOST_PROFILE})
-        conan_check(VERSION 1.24.0 REQUIRED DETECT_QUIET)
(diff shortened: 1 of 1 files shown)


a4z:
what is the status of this PR? 
nothing happens since more than 1 month. is there an other solution for using the new 2 profiles way, or something missing?

czoido:
Hi @christophschlosser thanks a lot for the contribution
Adding this to the milestone of the next release

cc @a4z

danimtb (approved):

jgsogo (changes requested):
IMHO, I would say that the `--profile:host` (or `--profile`, it is the same) should take the value from any of the existing profile variables: `DEBUG_PROFILE`, `RELEASE_PROFILE`, `RELWITHDEBINFO_PROFILE`, `MINSIZEREL_PROFILE` or `PROFILE` like it is doing right now, and we should **add only the new `BUILD_PROFILE` variable** (not sure if `PROFILE_BUILD` would be a better name).

conan.cmake
```diff
@@ -323,7 +331,7 @@ macro(parse_arguments)
   set(oneValueArgs CONANFILE  ARCH BUILD_TYPE INSTALL_FOLDER CONAN_COMMAND)
   set(multiValueArgs DEBUG_PROFILE RELEASE_PROFILE RELWITHDEBINFO_PROFILE MINSIZEREL_PROFILE
                      PROFILE REQUIRES OPTIONS IMPORTS SETTINGS BUILD ENV GENERATORS PROFILE_AUTO
-                     INSTALL_ARGS CONFIGURATION_TYPES)
+                     INSTALL_ARGS CONFIGURATION_TYPES BUILD_PROFILE HOST_PROFILE)
```
czoido, line 334:
```suggestion
                     INSTALL_ARGS CONFIGURATION_TYPES PROFILE_BUILD)
```

czoido:
Hi @cschlosser,
I have added some suggestions as I can't push to your branch. Could you please check when possible?

jgsogo (approved):

PR #252, opened 2020-06-23
Take CMAKE_CXX_FLAGS into account when detecting libcxx

robinlinden:
Hi! We're building for some obscure embedded platforms using sysroots and ran into an issue where conan fails to find string when trying to autodetect libcxx as it doesn't care about the CXX_FLAGS (where we pass --sysroot).

Just passing `${CMAKE_CXX_FLAGS}` doesn't work since the compiler will get "-m32 --sysroot=..." as 1 argument instead of two, so I had to split them using separate_arguments.

### DIFF
--- a/conan.cmake
+++ b/conan.cmake
@@ -264,9 +264,13 @@ function(conan_cmake_detect_unix_libcxx result)
         endif()
     endforeach()
 
+    # Take into account any compilation flags (e.g. --sysroot)
+    set(compile_flags ${CMAKE_CXX_FLAGS})
+    separate_arguments(compile_flags)
+
     execute_process(
         COMMAND ${CMAKE_COMMAND} -E echo "#include <string>"
-        COMMAND ${CMAKE_CXX_COMPILER} -x c++ ${compile_options} -E -dM -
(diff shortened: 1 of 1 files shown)


robinlinden:
If you want to be more selective in which flags are used, I'm fine with changing it so that it only grabs --sysroot from CMAKE_CXX_FLAGS.

smessmer:
This should probably also set the `libcxx` variable in `PARENT_SCOPE`, see https://github.com/conan-io/cmake-conan/issues/256

czoido:
Hi @robinlinden,
Thanks a lot for reporting and the contribution. I think this issue will be probably solved by https://github.com/conan-io/cmake-conan/pull/197/files# as well that is marked for review for version 0.16.
I'm going to close this as is duplicate but please feel free to reopen if you think it's necessary.

puetzk:
#197 doesn't parse it out of CXX_FLAGS (and I don't really think it should try), but it does handle sysroots if you set it the proper CMake way -- either `-D` or a toolchain which sets `CMAKE_SYSROOT` or `CMAKE_SYSROOT_COMPILE`/`CMAKE_SYSROOT_LINK` (if they need to be separate for some reason)

PR #253, opened 2020-06-25
Quote variables in conan_cmake_detect_vs_runtime

amerry:
Account for the possibility that variables are empty, or have values which are themselves variable names, in conan_cmake_detect_vs_runtime.

There are two issues this quoting resolves. One is errors resulting from empty variables - for exampe, if `CMAKE_CXX_FLAGS` is empty, the `string` call would have failed because it would have appeared to have too few arguments - quoting the final argument fixes this.

The other is that in the context of `if`, a lot of operators try to resolve their argument as a variable name first (but only if it's unquoted), meaning that `if(${build_type} STREQUAL "DEBUG")` would have unexpected results if `build_type` is set to `DEBUG` and there is a variable `DEBUG` set to `YES`, say. Quoting the variable expansion prevents this behaviour.


amerry:
As with #254 , this is not (yet) covered by a CLA, regardless of what CLAassistant says (although it's so trivial it's probably not covered by copyright anyway - that's my personal view, not my company's stance, though). I'll resubmit assuming I get the legal stuff sorted.

amerry:
Resurrecting - there were initially some CLA issues, but they're all resolved now and I can confirm that this code is covered by the CLA.

danimtb (approved):

memsharded:
Thanks for your contribution! (and sorry for the delay in reviewing it).
This will be included in next 0.16 release.

PR #254, opened 2020-06-25
Limit compile options considered for libcxx detection

amerry:
The COMPILE_OPTIONS directory property can contain generator
expressions, which are unlikely to be permitted by the compiler as valid
options. To limit the cases where this is an issue, only pull out the
options that might have an impact in conan_cmake_detect_unix_libcxx.

Note: we actually pull out -std= options as well as -stdlib options,
which may not be necessary, but is unlikely to be an issue.

### DIFF
--- a/conan.cmake
+++ b/conan.cmake
@@ -249,8 +249,15 @@ endfunction()
 
 
 function(conan_cmake_detect_unix_libcxx result)
+    set(compile_options)
+
     # Take into account any -stdlib in compile options
-    get_directory_property(compile_options DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} COMPILE_OPTIONS)
+    get_directory_property(options DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} COMPILE_OPTIONS)
+    foreach(option ${options})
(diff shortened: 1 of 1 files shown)


amerry:
I want to clarify that I messed up the CLA signing process - I did this on company time, and don't have the authority to sign the CLA, but I got confused by the CLA interface (ie: expected there to be a step *after* connecting my account, and wanted to see what it was). Hopefully I'll get the legal stuff sorted out and re-submit, but to be clear: this code is NOT covered by a CLA, no matter what CLAassistant thinks.

amerry:
Resurrecting - there were initially some CLA issues, but they're all resolved now and I can confirm that this code is covered by the CLA.

czoido:
Hi @amerry, 
Thanks a lot for reporting the issue and the contribution.
I think this issue is already managed by this PR: https://github.com/conan-io/cmake-conan/pull/197 that is marked for review for version 0.16, so if you don't mind I'll close the issue, please feel free to reopen if you think it's necessary.

amerry:
Yes, #197 deals with it much better in fact - I wasn't aware of string(GENEX_STRING).

Issue #255, opened 2020-06-29
conan_cmake_run(SETTINGS compiler.libcxx=libstdc++11) causes setting to be passed twice


smessmer:
I use the following lines to invoke conan:

```
conan_cmake_run(
       CONANFILE conanfile.py
       BUILD all
       SETTINGS compiler.libcxx=libstdc++11)
```

However, looking at the build output, I see

```
-- Conan executing: /usr/local/bin/conan install /root/cryfs/conanfile.py -s build_type=RelWithDebInfo -s compiler=gcc -s compiler.version=7 -s compiler.libcxx=libstdc++ -s compiler.libcxx=libstdc++11 -g=cmake --build
```

It seems, it just adds the `compiler.libcxx=libstdc++11` but still emits the `compiler.libcxx=libstdc++` and passes both to `conan install`.

Now it seems that in my case, this does the right thing and sets up the build to use `libstdc++11`, but this seems fishy nevertheless and I'm not sure if I can rely on that behavior.

czoido:
Hi @smessmer,

Thanks a lot for reporting. Yes, it seems like the autodetection of `libcxx` is run regardless of the arguments provided and for this case the `SETTINGS` defined in `conan_cmake_run` should take preference. 
I'll add this fix for the next release.

PR #258, opened 2020-06-30
Remove manually specified settings from auto list

czoido:
Manual settings were added in the end to automatic CMake settings, so they were not taken into account because only the first one was considered by CMake.
This PR removes the MANUAL settings from the AUTO list of added settings from CMake so that they are not added twice.

Fixes: #255

### DIFF
--- a/conan.cmake
+++ b/conan.cmake
@@ -229,6 +229,13 @@ function(conan_cmake_settings result)
                                    compiler.runtime compiler.libcxx compiler.toolset)
     endif()
 
+    # remove any manually specified settings from the autodetected settings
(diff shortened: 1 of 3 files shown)


conan.cmake
```diff
@@ -229,6 +229,13 @@ function(conan_cmake_settings result)
                                    compiler.runtime compiler.libcxx compiler.toolset)
     endif()
 
+    # remove any manually specified settings from the autodetected settings
+    foreach(ARG ${ARGUMENTS_SETTINGS})
```
memsharded, line 233: A test would be great.

memsharded (approved):

PR #280, opened 2020-09-13
Support OPTIONS with REQUIRES, not just with CONANFILE

### LINKED ISSUE #279: `OPTIONS` only works with `CONANFILE`, not with `REQUIRES`
It appears that the support for `OPTIONS` is nested inside of the support for `CONANFILE`
https://github.com/conan-io/cmake-conan/blob/2c98e07189eb62bf103cefa603f16d343ee4d75d/conan.cmake#L409
https://github.com/conan-io/cmake-conan/blob/2c98e07189eb62bf103cefa603f16d343ee4d75d/conan.cmake#L415-L418

It is not clear to me what's wrong with setting options for packages listed in `REQUIRES`, so presumably this should just be outside the if?

puetzk:
Fixes #279


puetzk:
Rebased to develop, oops.

puetzk:
Updated per @czoido at https://github.com/conan-io/cmake-conan/issues/279#issuecomment-693203626.
If we consistently pass `OPTIONS` on the command-line (so that they override anything from a profile) writing them into the conanfile.txt too is redundant.

czoido (approved):

czoido:
Thanks a lot for the contribution. I added a test. LGTM!

PR #285, opened 2020-10-09
Update conan.cmake

redradist:
Added Generic platform for embedded development

### DIFF
--- a/conan.cmake
+++ b/conan.cmake
@@ -106,7 +106,7 @@ function(conan_cmake_settings result)
         if(${CMAKE_SYSTEM_NAME} STREQUAL "QNX")
             set(CONAN_SYSTEM_NAME Neutrino)
         endif()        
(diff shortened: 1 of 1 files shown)


czoido:
Hi @redradist,
I'm thinking that maybe it would be a better solution to skip the platform check when `CMAKE_SYSTEM_NAME` is set to `Generic` instead of adding `Generic`.
Something like: https://github.com/czoido/cmake-conan/blob/351f6b77a4e7af9f5af4713a872f94e1baf013d8/conan.cmake#L100-L117

redradist:
> Hi @redradist,
> I'm thinking that maybe it would be a better solution to skip the platform check when `CMAKE_SYSTEM_NAME` is set to `Generic` instead of adding `Generic`.
> Something like: https://github.com/czoido/cmake-conan/blob/351f6b77a4e7af9f5af4713a872f94e1baf013d8/conan.cmake#L100-L117

@czoido Maybe you are right, because then somebody can set name to `NewFreeRTOS` and script should be fixed again

czoido:
Hi @redradist,
Could you please update the PR with the changes we discussed and sign the License Agreement so we can put this under review for 0.16 release?

czoido:
Hi, closing this as this problem should be solved by https://github.com/conan-io/cmake-conan/pull/292

PR #293, opened 2020-10-29
Initial support for architecure detection

redradist:


redradist:
@czoido Could you review this changes ? It seems like very useful as for, because it allow auto-detect architecture without manually specifying it ;)

redradist:
@czoido Could you review this pull-request, I've fixed merged conflicts ... It helps detect architecture by detecting prefix of compiler

czoido:
Thanks for contributing, I'm closing this as this is outdated, not Conan supports xtensa architectures: https://github.com/conan-io/conan/blob/bd78684d13993d113a591240f1913ed4b7d03390/conans/client/conf/__init__.py#L82
And it can be set in the settings.

redradist:
@czoido I know about that xtensa was added, it was added by me :)))

https://github.com/conan-io/conan/pull/7977

But this PR is still relevant because it is responsible for auto detecting of system settings:
```cmake
conan_cmake_autodetect(settings)
```

Why do you think it is not needed ?

redradist:
@czoido What is the issue with this PR ?
It is responsible for auto detection settings, exactly what is responsible `conan_cmake_autodetect(settings)`

czoido:
Hi @redradist,
I don't think we should add this autodetection for these specific cases. If you want to use that architecture you can pass the arch setting or the profile argument instead of relying on cmake-conan to do that.

Issue #308, opened 2021-02-13
conan.cmake fails when using pyenv


BlueSolei:
## Issue
Hi,
The conan.cmake script fails due to the `conan` command is not an executable.
I use [PyEnv](https://github.com/pyenv-win/pyenv-win) on windows, so my `conan` command is a .bat file, or if running in Git-Bash, its a sh script.

The error I get is `Conan install failed='%1 is not a valid Win32 application`
Here is the error:
```shell
√ /c/dev/expr/conan-training/requires_conditional (shaul) $ cmake -B build -S src/
-- Building for: Visual Studio 16 2019
-- The CXX compiler identification is MSVC 19.27.29110.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Professional/VC/Tools/MSVC/14.27.29110/bin/Hostx64/x64/cl.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Conan: Using cmake-multi generator
-- Conan: Automatic detection of conan settings from cmake
-- Conan: Detected VS runtime: MD
-- Conan: Settings= -s;arch=x86_64;-s;build_type=Release;-s;compiler=Visual Studio;-s;compiler.version=16;-s;compiler.runtime=MD
-- Conan: checking conan executable
-- Conan: Found program C:/Users/sfridman/.pyenv/pyenv-win/shims/conan
-- Conan: Version found
-- Conan executing: C:/Users/sfridman/.pyenv/pyenv-win/shims/conan install . -s arch=x86_64 -s build_type=Release -s compiler=Visual Studio -s compiler.version=16 -s compiler.runtime=MD -g=cmake_multi
CMake Error at C:/dev/expr/cmake-conan/conan.cmake:455 (message):
  Conan install failed='%1 is not a valid Win32 application'
Call Stack (most recent call first):
  C:/dev/expr/cmake-conan/conan.cmake:540 (conan_cmake_install)
  CMakeLists.txt:5 (conan_cmake_run)


-- Configuring incomplete, errors occurred!
See also "C:/dev/expr/conan-training/requires_conditional/build/CMakeFiles/CMakeOutput.log".
```
## A fix
I manage to fix the issue by running the calls to `conan` in a shell, like this:
 ```cmake
set(MY_CONAN_CMD "cmd.exe")
[...]

PR #309, opened 2021-02-13
Check & log conan commands execution

BlueSolei:
After having issues with the script failed to run `conan install`, [due to the use of pyenv](https://github.com/conan-io/cmake-conan/issues/308) I searched to see why other commands didn't fail and realized that they did fail, but didn't report it.
So added here the check & report code.


czoido:
Hi @BlueSolei,
Thanks a lot for your contribution, we'll add this check for the next release.

czoido (approved):

PR #319, opened 2021-03-02
Add a function to call `conan lock create`

prince-chrismc:
Follow up of #294

The new workflow is really good 👏 The only challenge was the base-lockfile workflow is not supported. So I modified the new `conan_cmake_install` to enable this feature.


czoido:
Hi @prince-chrismc,
Thanks a lot for the contribution. Could you please give some details of the flow you want to follow using `conan_cmake_lock_create`?

prince-chrismc:
I am using the base lockfile workflow. https://github.com/conan-io/docs/issues/1904

I build many different configuration so, I only have a base lockfile to make sure all the revisions are pinned and easily managed.
https://github.com/prince-chrismc/user-management/blob/master/backend/conan.lock

In order for this to be usable with the install command, I need to create a full lock for that profile/settings/options
This extra step is what I am trying to accomplish...

Here's the implementation I am using as a proof of concept https://github.com/prince-chrismc/user-management/blob/master/backend/cmake/conan-setup.cmake

czoido:
Hi @prince-chrismc,
Thanks a lot and sorry for the delay. I think this is a valid feature and we can add it for the next release.
I'll add some documentation and also tests to check the feature.

czoido (approved):

dgoel:
This change looks great and I'm looking forward to using it. I've one question though.

Assuming both `conanfile.txt` (or `conanfile.py`) and `base_conan.lock` files are tracked by the version control in a repo, then there's a scenario where reproducibility feature of lockfile is lost. For example, a developer updates `conanfile.txt` to add a new package `packageXYZ` but forgets to update the `base_conan.lock` file. As the complete lock file is generated using the proposed `conan_cmake_lock_create` function, the underlying `conan lock create` command would happily work and only throw the following warning:
```
conanfile.txt: WARN: Require 'packageXYZ' cannot be found in lockfile
```

Is this the expected behavior? It would be really nice if the build actually fails in this case and requires the developer to manually update the base lock file to pin all the dependencies. I couldn't find a command line option to do that though. Directly using `conan_cmake_install` does throw this error but requires the inclusion of compiler settings etc. in the lockfile (which is less than ideal).

Any suggestions on how to handle this scenario?

prince-chrismc:
@czoido I resolved some merge conflicts. it would be great to merge this :pray:

SvenRoederer:
@dgoel The point you mentioned is worth keeping in mind. But the change you mention is not part of this PR, right? If not, was it PRed in some other way?

memsharded:
Hi @SvenRoederer 

This thread was quite outdated, referring to the legacy ``cmake-conan`` integration that has been superseded by the new one for Conan 2 with CMake dependency providers. You might want to create new tickets instead, thanks for the feedback.

Issue #322, opened 2021-03-04
[Windows] Macro _conan_detect_build_type doesnt work with Visual Studio IDE


Sneder89:
Hello everybody!

when I invoke _conan_cmake_autodetect_ with Visual Studio IDE, the function returns inside the macro __conan_detect_build_type_  an error message because _CMAKE_BUILD_TYPE_ is not defined. Multi Configuration IDE's like VS or XCode doesn't set the Variable _CMAKE_BUILD_TYPE_ but _CMAKE_CONFIGURATION_TYPES_. 

A solution could be that the Variable _CMAKE_CONFIGURATION_TYPES_ is getting evaluated too after _CMAKE_BUILD_TYPE_ and the corresponding settings are getting adapted.

Nekto89:
I have the same issue. Had to switch to conan_cmake_run

czoido:
Hi @Sneder89, @Nekto89,
Thanks a lot for reporting the issue, I have reproduced it. We'll try to fix this as soon as possible.

czoido:
Hi @Sneder89, @Nekto89 

Having a more close look at the issue, I think this is how it should work and the _CMAKE_CONFIGURATION_TYPES_ and `CMAKE_BUILD_TYPE` should be handled from the consumer _CMakelists.txt_.
In the old flow using `conan_cmake_run` for Multi Configuration generators would call several times to `conan install` but as now we are decoupling and separating this behaviours and make it more aligned with conan commands calls we can't handle this from the inside of `conan.cmake`.
I have added a test to check this works: https://github.com/conan-io/cmake-conan/pull/325/files#diff-f455f302936271d755f6892d443786094d2607b505ff180ca817cb9eeabe1e5eR941-R964
Will this help to solve the problem?

Sneder89:
Hi @czoido,

this helps and solves my problem. But I agree that its complicated to have manually set `CMAKE_BUILD_TYPE` to use the `conan_settings_autodetect` function.

czoido:
Hi @Sneder89,
I have made some changes so that `conan_cmake_autodetect` cane be called with the `BUILD_TYPE` argument so it's more explicit.

PR #325, opened 2021-03-15
Testing new flow for Multi Configuration generators

czoido:
Closes: https://github.com/conan-io/cmake-conan/issues/322


puetzk:
I think it needs to be made clearer that naming the loop variable `CMAKE_BUILD_TYPE` was important - the fact that's a reserved special cmake-variable name is the only reason it has any effect on conan_cmake_autodetect, which is rather subtle. And using those reserved names as local variables is not usually a good practice, so that's kind of sneaky.

czoido:
> I think it needs to be made clearer that naming the loop variable `CMAKE_BUILD_TYPE` was important - the fact that's a reserved special cmake-variable name is the only reason it has any effect on conan_cmake_autodetect, which is rather subtle. And using those reserved names as local variables is not usually a good practice, so that's kind of sneaky.

Hi @puetzk,
Thanks for the comments, I think you are right, I have introduced some changes to make it more explicit, passing the `BUILD_TYPE` to `conan_cmake_autodetect` so that the use of the CMake reserved variable does not lead to confusion.

memsharded (approved):

PR #327, opened 2021-03-22
Remove duplicate include and modernize example

stephan-cr:


czoido:
Hi  @stephan-cr,
Thanks a lot for the contribution.
The duplicated include is ok, it is supposed to include the downloaded `conan.cmake` in case someone uncomments that block of code.
Also, the cmake_minimum_required should be raised to 3.8 when `cxx_std_11` was introduced.

stephan-cr:
@czoido Resolved the merge conflict.

I would also propose to include `conan.cmake` like `include(conan)`, because that's the point of having `list(APPEND CMAKE_MODULE_PATH ${CMAKE_BINARY_DIR})` at the top.

PR #335, opened 2021-04-14
Add conan_add_user function

gomons:
conan.cmake has function to add new remote but doesn't have function to add user for this remote.


gomons:
Ok, now I see this function is useless because conan supports CONAN_LOGIN_USERNAME and CONAN_PASSWORD environment variables (=

PR #337, opened 2021-04-27
support conan installed using pip --user flag

Renari:


Renari:
An already supported solution to this problem is specifying the `CONAN_COMMAND` argument.

PR #343, opened 2021-06-22
Detect compiler.cppstd from CMAKE_CXX_STANDARD

guy-david:
Conan dependencies do not share the same CMAKE_CXX_STANDARD set by a user in a parent project.


guy-david:
ping.

guy-david:
That's true, however this change is not a feature.
The current default behaviour can build dependencies with the wrong C++ standard level, which are not all ABI compatible, leading to a segmentation fault.

PR #353, opened 2021-08-15
Add conan_imports function

maksim-0:
There is NO_IMPORTS argument in conan_cmake_install and we use it in our projects. Then in order to copy artifacts to specific import folder we need to call `conan imports`.

### DIFF
--- a/conan.cmake
+++ b/conan.cmake
@@ -900,3 +900,65 @@ macro(conan_config_install)
     message(FATAL_ERROR "Conan config failed='${return_code}'")
   endif()
 endmacro()
+
+function(conan_imports)
+    # Calls "conan imports"
+    # Argument PATH is required, INSTALL_FOLDER, IMPORT_FOLDER and UNDO are optional
+    # Example usage:
+    #    conan_imports(PATH . IMPORT_FOLDER ${IMPORT_FOLDER})
+    if(DEFINED CONAN_COMMAND)
+        set(CONAN_CMD ${CONAN_COMMAND})
+    else()
+        conan_check(REQUIRED)
+    endif()
+
+    set(options UNDO OUTPUT_QUIET ERROR_QUIET)
+    set(oneValueArgs PATH INSTALL_FOLDER IMPORT_FOLDER)
+    cmake_parse_arguments(ARGS "${options}" "${oneValueArgs}" "" ${ARGN})
+
+    foreach(arg ${options})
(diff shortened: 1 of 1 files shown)


czoido:
Hi @maksim-0,
Thanks a lot for the contribution. Please, could you provide a little bit more background on why you do a `conan install` with -`--no-imports` for later calling `conan imports `? Shouldn't the result be the same than doing the imports in the `conan install` ?

maksim-0:
@czoido 

I was using it like this:
```cmake
# PRODUCT_BUILD_DIR is outside CMAKE_BINARY_DIR
set(
    CMAKE_RUNTIME_OUTPUT_DIRECTORY
    ${PRODUCT_BUILD_DIR}/bin
)
set(
    CMAKE_LIBRARY_OUTPUT_DIRECTORY
    ${PRODUCT_BUILD_DIR}/lib
)
set(
    CMAKE_ARCHIVE_OUTPUT_DIRECTORY
    ${PRODUCT_BUILD_DIR}/lib
)

conan_cmake_configure(
    REQUIRES
        boost/1.72.0
    IMPORTS
        "bin, *.dll -> ./bin"
        "lib, *.so* -> ./lib"
        "licenses, LICENSE_1_0.txt -> ./licenses @ root_package=boost, folder=True"
)

conan_cmake_autodetect(CONAN_SETTINGS)

conan_cmake_install(
    PATH_OR_REFERENCE .
    BUILD never
    SETTINGS ${CONAN_SETTINGS}
)

conan_imports(
    PATH .
    IMPORT_FOLDER ${PRODUCT_BUILD_DIR}
)
```
But you raised a good question. I was absolutely sure that I can't specify absolute paths in IMPORTS "section" of  ``conan_cmake_configure``. I tried and it appears that I actually can. So I removed ``conan_imports`` and rewrote ``conan_cmake_configure``:
```cmake
conan_cmake_configure(
    REQUIRES
        boost/1.72.0
    IMPORTS
        "bin, *.dll -> ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}"
        "lib, *.so* -> ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}"
        "licenses, LICENSE_1_0.txt -> ${PRODUCT_BUILD_DIR}/licenses @ root_package=boost, folder=True"
)
```
I'm sorry for taking up your time. We don't really need ``conan_imports``.

Issue #32, opened 2017-05-14
conan_cmake_run fails when invoked via CLion


sztomi:
CMake:

```cmake
if(NOT EXISTS "${CMAKE_BINARY_DIR}/conan.cmake")
    message(STATUS "Downloading conan.cmake from https://github.com/memsharded/cmake-conan")
    file(DOWNLOAD "https://raw.githubusercontent.com/conan-io/cmake-conan/master/conan.cmake"
            "${CMAKE_BINARY_DIR}/conan.cmake")
endif()

include(${CMAKE_BINARY_DIR}/conan.cmake)
conan_cmake_run(CONANFILE conanfile.txt  # or relative build/conanfile.txt
                BASIC_SETUP CMAKE_TARGETS
                BUILD missing)
```

Output:

```
-- Conan ** WARNING** : This detection of settings from cmake is experimental and incomplete. Please check 'conan.cmake' and contribute
-- Conan executing: conan install -f=/home/sztomi/Projects/MyCoolProject/conanfile.txt -g cmake -s build_type=Debug -s os=Linux -s compiler=gcc -s compiler.version=6.3 -s compiler.libcxx=libstdc++11 --build=missing
CMake Error at cmake-build-debug/conan.cmake:221 (message):
  Conan install failed='No such file or directory'
Call Stack (most recent call first):
  cmake-build-debug/conan.cmake:298 (conan_cmake_install)
  CMakeLists.txt:14 (conan_cmake_run)
```

However, when I run the same command from the command line, it works. Tried CMake 3.7 (bundled) and 3.8 with the same result.

sztomi:
Interestingly, it fails the same way in KDevelop. My guess is that the `conan` executable is not found because I don't have it system-wide installed (I installed via `pip install conan --user`, so it lives in ~/.local/bin)

sztomi:
Yep, after symlinking conan to /usr/bin, it works without a hitch.

memsharded:
I think ``cmake-conan`` has a configurable ``CONAN_COMMAND`` where you could provide the full path to conan executable. Is there any other suggestion of anything we could fix? Not sure what we could do to improve this issue.

sztomi:
Sorry I wasn't very explicit about this, but ~/.local/bin is in my $PATH (set in .bashrc), so I would expect the default `conan` command to work (without an absolute path). As mentioned, if I copy the logged command line to my terminal it works; and I think it should from cmake as well.

memsharded:
[...]

PR #359, opened 2021-09-02
add PATHS ~/.local/bin

Sebanisu:
Default on ubuntu is to install conan into ~/.local/bin and cmake won't look there unless you give it a hint.


Elnee:
Need it badly) I think that's why I have conan check error.

Elnee:
@Sebanisu, according to this documentation maybe it's better to change `HINTS` to `PATHS`.

> Search the paths specified by the HINTS option. These should be paths computed by system introspection, such as a hint provided by the location of another item already found. Hard-coded guesses should be specified with the PATHS option.

Source: [find_program - CMake 3.22.0 Documentation](https://cmake.org/cmake/help/latest/command/find_program.html)

czoido:
Hi @Sebanisu, @Elnee

You can set the [CONAN_COMMAND](https://github.com/conan-io/cmake-conan#conan_command) variable in your CMakeLists.txt for this:

```
set(CONAN_COMMAND "/usr/.local/bin/conan")
```

I'll close this, as that is the recommended way of setting a custom location for Conan, please feel free to reopen or open an issue if you have more comments or this does not work for you. Thanks a lot.

Sebanisu:
I'm having no success with [CONAN_COMMAND](https://github.com/conan-io/cmake-conan#conan_command)  I also don't like hard coding the location of Conan when depending on the system it might be installed in a different place. `~/.local/bin/conan` is where pip installed it. I put that path in and it just failed to find it. It also caused the systems that were working to break.

Sebanisu:
https://github.com/conan-io/cmake-conan/issues/32 there are some solutions in this issue.

PR #374, opened 2021-11-30
Added conan.cmake file version check & download error reporting

jlschrag:
The current version of the conan.cmake file download in the readme is just a simple check that the file exists.  This is problematic if a developer has an outdated version of the file present on their machine and can result in vague error messages later in the process.  I've added logic to parse the version number from the file present on their machine and compare with the version in their CMakeLists.txt.

I've also added logic to catch file download errors as errors were previously either ignored or a misleading error message was shown.

Finally, I've updated the conan.cmake file version number to 0.17.0 avoid errors for Visual Studio 2022 users following the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,9 +29,9 @@ add_definitions("-std=c++11")
 
 if(NOT EXISTS "${CMAKE_BINARY_DIR}/conan.cmake")
   message(STATUS "Downloading conan.cmake from https://github.com/conan-io/cmake-conan")
-  file(DOWNLOAD "https://raw.githubusercontent.com/conan-io/cmake-conan/v0.16.1/conan.cmake"
+  file(DOWNLOAD "https://raw.githubusercontent.com/conan-io/cmake-conan/0.17.0/conan.cmake"
                 "${CMAKE_BINARY_DIR}/conan.cmake"
-                EXPECTED_HASH SHA256=396e16d0f5eabdc6a14afddbcfff62a54a7ee75c6da23f32f7a31bc85db23484
+                EXPECTED_HASH SHA256=3bef79da16c2e031dc429e1dac87a08b9226418b300ce004cc125a82687baeef
                 TLS_VERIFY ON)
 endif()
 
--- a/README.md
+++ b/README.md
@@ -27,12 +27,35 @@ list(APPEND CMAKE_PREFIX_PATH ${CMAKE_BINARY_DIR})
 
 add_definitions("-std=c++11")
 
+set (CONAN_CMAKE_VERSION "0.17.0")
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -27,12 +27,35 @@ list(APPEND CMAKE_PREFIX_PATH ${CMAKE_BINARY_DIR})
 
 add_definitions("-std=c++11")
 
+set (CONAN_CMAKE_VERSION "0.17.0")
```
jlschrag, line 30: Note: This & the logic below assume that the format of the file version will be the same going forward within the file _and_ in the file download URL.  This historically has not always been the case.  Version 0.16 appears as "0.16.0" in the file, but "0.16" in the URL.

README.md
```diff
@@ -27,12 +27,35 @@ list(APPEND CMAKE_PREFIX_PATH ${CMAKE_BINARY_DIR})
...
+set (CONAN_CMAKE_HASH 3bef79da16c2e031dc429e1dac87a08b9226418b300ce004cc125a82687baeef)
+
+set (DOWNLOAD_CONAN_CMAKE false)
+
 if(NOT EXISTS "${CMAKE_BINARY_DIR}/conan.cmake")
+  set (DOWNLOAD_CONAN_CMAKE true)
+else()
+  message(STATUS "Checking version of existing conan.cmake file")
+  file(READ "${CMAKE_BINARY_DIR}/conan.cmake" FILE_CONTENTS)
+  string(FIND "${FILE_CONTENTS}" "# version: ${CONAN_CMAKE_VERSION}" MATCH_RESULT)
```
prince-chrismc, line 40: If you check the hash you should be fine?

prince-chrismc (commented):
I think this is worth adding since everyone needs to copy paste it... perhaps in a separate example? 

https://github.com/prince-chrismc/user-management/blob/9df3631cc2bdedaf3df136cbdf0d77f60bdf796a/backend/cmake/conan-setup.cmake#L6

README.md
jlschrag: Good call!

Turns out the file download command checks the hash of an existing file before downloading.  If the file already exists, it returns the message "returning early; file already exists with expected SHA256 hash".  I've updated the example.

jlschrag:
@prince-chrismc Are you saying the CONAN_WRAPPER definition should be added elsewhere in the README?

prince-chrismc:
Yes a personal preference, the first example should be easy/most simple.

There's no need to make any changes.

jlschrag:
@prince-chrismc Is this waiting for me to add CONAN_WRAPPER before merge?

prince-chrismc:
I am not on the Conan team, and they are very busy with the near v2 release.

Last time I said "no needed for changes" so we just need to wait for them to attack this project as well :)

PR #376, opened 2021-12-06
feat: allow users to specify conan executable for conan_config_install

charles-huet-ixxi:
In my use case I need to specify the conan executable when installing the config.

This fixes my workflow while being coherent with the rest of the file.


czoido (approved):

czoido:
Hi @charles-huet-ixxi thanks a lot for contributing this! It will be released in 0.18

Issue #385, opened 2022-01-07
Conan not detected when installed through winget


LtdSauce:
I installed conan using the `winget` tool on my windows 10 machine. When i use the cmake files in a project to detect conan it then fails with the message, that cmake cannot find the conan program. 
This seems to be due to the fact, that `winget` installs conan as `Conan` instead of `conan`. This lets the `find_program` call in the `conan_check` macro fail.

The behavior can be reproduced using the [cpp_starter_project](https://github.com/cpp-best-practices/cpp_starter_project) after installing conan through the command `winget install "Conan Package Manager"`.

The problem could be solved with changing Line 800 to also check for `Conan` in addition to `conan`

LtdSauce:
As already mentioned in cpp-best-practices/cpp_starter_project#179 the problem has disappeared after a regular reboot. 
I'll close this issue and the related PR as i don't think it ever was a real issue.

PR #386, opened 2022-01-07
Add Conan as a possible program name in conan_check

LtdSauce:
This commit extends the `find_program` to also look for `Conan` in addition to `conan`.
The reason for this change is, that conan is named `Conan` when installed with winget.


LtdSauce:
This should solve #385

LtdSauce:
I close this PR. See #385 for explanation.

PR #407, opened 2022-04-01
Fix has for 0.18.0 release.

AlphaCluster:
Looks like the readme's hash wasn't updated for the 0.18.0 release.


czoido:
Hi @AlphaCluster,
Thanks a lot for the contribution. We'll make a patch for 0.18 and release today the fix.

czoido:
This has to be fixed in the release/0.18 branch, not to be target to develop, I'll make the changes to target the other branch
UPDATE: I'll merge this for the 0.18.1 release but removing the check of the hash, that is something that will bite again and should be a choice taken by the developer that uses conan.cmake

PR #420, opened 2022-05-06
Restore detection of target OS

saierd:
This was broken in 61ae3b3356f2f2c6ca2a3dd0fedcc21a407bbe1c. The variable is getting set correctly, but it was never added to the actual settings.

### DIFF
--- a/conan.cmake
+++ b/conan.cmake
@@ -280,7 +280,7 @@ function(conan_cmake_settings result)
 
     if(NOT _SETTINGS OR ARGUMENTS_PROFILE_AUTO STREQUAL "ALL")
         set(ARGUMENTS_PROFILE_AUTO arch build_type compiler compiler.version
-                                   compiler.runtime compiler.libcxx compiler.toolset)
+                                   compiler.runtime compiler.libcxx compiler.toolset os)
     endif()
 
     # remove any manually specified settings from the autodetected settings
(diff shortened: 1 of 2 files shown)


tk-tnz:
@czoido I just had the same issue, not getting the os from `conan_cmake_autodetect` - the workaround is easy but it would be great if this PR could be merged. Thanks in advance.

czoido:
> @czoido I just had the same issue, not getting the os from `conan_cmake_autodetect` - the workaround is easy but it would be great if this PR could be merged. Thanks in advance.

Hi @tk-tnz,

Thanks for the ping and sorry for the late response. We are focusing our efforts in the `develop2` branch that's compatible with Conan 2.0 and has a much consistent approach using CMake's dependency providers. Pleas, have a look if you want to continue using cmake-conan with newer Conan versions. 

Anyway, I have added a test for this and personally think that it could be merged to consider cases that do not have the `os` defined in the default profile (as a temporary workaround you may define it there and it should work). I would be ok merging this but I'm assigning @memsharded this one so he can take the final decision in case there's something I'm missing.

memsharded:
We would need @saierd re-signing the CLA, there were some changes some time ago that requires re-signing.
Could you please check that @saierd?

tk-tnz:
> Thanks for the ping and sorry for the late response. We are focusing our efforts in the `develop2` branch that's compatible with Conan 2.0 and has a much consistent approach using CMake's dependency providers. Pleas, have a look if you want to continue using cmake-conan with newer Conan versions.

Thanks for your prompt replies! Yes, I looked at the develop2 branch but haven't had the time to look at how feasible it will be to migrate my existing setup (which is not the most trivial with conan being automatically invoked by cmake a number of times with dynamically generated profiles from cmake in different subdirectories).

saierd:
> We would need @saierd re-signing the CLA, there were some changes some time ago that requires re-signing. Could you please check that @saierd?

I have signed the CLA again.

memsharded (approved):

PR #429, opened 2022-05-25
Improve cmake detection

buresu:
In some environments there is no path to the cmake.
Because the CMAKE_COMMAND variable contains the path to the currently used cmake, set it to the CONAN_CMAKE_PROGRAM will solve this problem.
Please see https://github.com/conan-io/conan/issues/7147

### DIFF
--- a/conan.cmake
+++ b/conan.cmake
@@ -524,6 +524,7 @@ function(old_conan_cmake_install)
     string (REPLACE ";" " " _conan_args "${conan_args}")
     message(STATUS "Conan executing: ${CONAN_CMD} ${_conan_args}")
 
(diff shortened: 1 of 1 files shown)


czoido:
Hi @buresu,
Thanks for the contribution but I'm afraid we can't merge this, it's modifying the `conan_cmake_run()` function that is going to be deprecated soon and also will overwrite `CONAN_CMAKE_PROGRAM` that someone could be using. I think this issue is probably more related to something that has to be fixed in the build environment than something to fix in cmake-conan.

buresu:
@czoido Hi, Thank you for the review.
I understood that, I will close this PR.

PR #433, opened 2022-06-13
improve conan generator choice

pgrohs:
in case the cmake_find_package generator is present in the recipe, the script should enable also the cmake_find_package_multi generator in case of multi-config projects


czoido:
Hi @pgrohs,
Sorry for the late review, and thanks for the contribution. I'm afraid this this is modifying part of the `conan_cmake_run` implementation that we don't recommend any more, please check the recommended way of using cmake-conan using `conan_cmake_install`

Issue #333, opened 2021-04-07
How to use conan_config_install with --args argument


gomons:
Usage example of **conan config install** command (install conan config from git repository on tag v1.0):
`conan config isntall git@github.com:gomons/conan_config.git --args="-b v1.0"`

How to call the command above by using **conan_config_install** macro?

To make it work I need to remove extra quotes in **conan_config_install** implementation:
`set(CONAN_ARGS_ARGS "--args=\"${CONAN_ARGS}\"")`
change to this:
`set(CONAN_ARGS_ARGS "--args=${CONAN_ARGS}")`

It's also unclear to me why **ARGS** is a multi value argument and not single value argument with string value.

PR #443, opened 2022-08-29
conan config install uses the args as multiple arguments and passes them correctly to git

martin4861:
Hi, 
I slightly changed the behavior of `conan_config_install()`, when using ARGS. This problem is also mentioned in https://github.com/conan-io/cmake-conan/issues/333.
`conan config install()` uses now a space seperated list and passes it correctly to conan/git:
e.g. `conan_config_install(ITEM https://github.com/conan-io/cmake-conan.git
                                 TYPE git
                                 ARGS -b v0.5)`
Moreover I introduced a unit test, which shows the intended behavior.

Please discuss, if this change is okay. 

Thank you,
Martin

### DIFF
--- a/tests.py
+++ b/tests.py
@@ -831,6 +831,29 @@ class CMakeConanTest(unittest.TestCase):
             assert remote["url"] == remote_url, "Invalid remote url"
             assert remote["verify_ssl"] == verify_ssl, "Invalid verify_ssl"
 
+    def test_conan_config_install_args(self):
+        content = textwrap.dedent("""
+            cmake_minimum_required(VERSION 3.9)
+            project(FormatOutput CXX)
+            message(STATUS "CMAKE VERSION: ${CMAKE_VERSION}")
+
+            set(CONAN_DISABLE_CHECK_COMPILER ON)
+            include(conan.cmake)
+            conan_config_install(ITEM https://github.com/conan-io/cmake-conan.git
+                                 TYPE git
+                                 ARGS -b v0.5)
+        """)
+        save("CMakeLists.txt", content)
+        os.makedirs("build")
+        os.chdir("build")
+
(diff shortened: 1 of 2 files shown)


martin4861:
Hello @memsharded or @czoido, may I ask one of you for a review and some discussions on this PR? Thanks a lot!

tests.py
```diff
@@ -831,6 +831,26 @@ def test_conan_config_install(self):
...
+            conan_config_install(ITEM https://github.com/conan-io/cmake-conan.git
+                                 TYPE git
+                                 ARGS -b v0.5)
+        """)
+        save("CMakeLists.txt", content)
+        os.makedirs("build")
+        os.chdir("build")
+
+        os.environ.update({"CONAN_TRACE_FILE": f"{os.getcwd()}/trace.log", "CONAN_VERBOSE_TRACEBACK": "1", "CONAN_LOGGING_LEVEL" : "40"})
+
```
czoido:
```suggestion
```
I think we can remove this

tests.py
```diff
@@ -831,6 +831,26 @@ def test_conan_config_install(self):
...
+        run("cmake .. %s" % generator)
```
czoido: Maybe you can check something in the output to validate that the conan config install was succesful like in other tests:

```python
        run("cmake .. {} -DCMAKE_BUILD_TYPE=Release > output.txt".format(generator))
        with open('output.txt', 'r') as file:
            data = file.read()
            assert "Repo cloned!" in data
            assert "Copying file" in data
```

czoido:
Hi @martin4861,

Thanks a lot for the contribution and sorry for the delay, it's looking good but check my comments

tests.py
martin4861: Perfect, I'll take that over :+1:

martin4861:
Hi @czoido, 
thanks for your review. I updated the test as suggested.

czoido (approved):

PR #446, opened 2022-09-24
Add 'conan_version()' function to get the Conan client version

hwhsu1231:
Add `conan_version()` function to aquire the current Conan version.

Since some of `settings` are introduced in Conan 2.X but not in Conan 1.X or some of them are deprecated in Conan 2.X, for example when using MSVC compiler,

- In **Conan 1.X**, we may use `Visual Studio` in `compiler`.
- In **Conan 2.X**, we need to use `msvc` in `compiler`. (because `Visual Studio` is removed in `settings.yml`)

Therefore, I think `conan-cmake` needs a mechanism to determine whether it is using **Conan 1.X** or **Conan 2.X**.

The following is an example usage of `conan_version()` function.

```cmake
conan_version(CONAN_VERSION)
if(CONAN_VERSION VERSION_GREATER_EQUAL 2.0.0)
    # For Conan 2.X
else()
    # For Conan 1.X
endif()
```

### DIFF
--- a/.github/workflows/cmake_conan.yml
+++ b/.github/workflows/cmake_conan.yml
@@ -21,6 +21,6 @@ jobs:
       - name: Install Conan
         run: pip install conan pytest && conan --version
       - name: Install CMake
-        run: conan install cmake/3.16.4@ -g=virtualrunenv
+        run: conan install cmake/3.16.9@ -g=virtualrunenv
       - name: Run Tests
         run: ${{matrix.ENV_CMAKE}} && pytest tests.py -rA -v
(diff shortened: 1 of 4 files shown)


hwhsu1231:
@czoido Please check it. Thank you.

czoido:
Hi @hwhsu1231,

Thanks a lot for the contribution, I think that the feature is valid, but let me do some refactors and add a test and the docs. 
I changed the name to `conan_version` and also called that inside `conan_check` to reuse the functionality, also added a test.

hwhsu1231:
@czoido There's a typo on `README.md` in this merged PR.

image

czoido:
Oh, sorry, thanks a lot, I'll fix it

PR #451, opened 2022-10-22
Add 'conan_cmake_profile' function and its test

### LINKED ISSUE #450: [feature] Design a function to generate required profiles, `conan_cmake_profile()`.
### Description

In my opinion, the most essential commands of cmake-conan for now are:

- `conan_cmake_autodetect()`: Collect settings which will be passed to `conan_cmake_install(SETTINGS)`.
- `conan_cmake_install()`: Running the `conan install` command.

Since it is encouraged to use **Host Profile** and **Build Profile** when running `conan install` command in the future and [Conan 2.X](https://docs.conan.io/en/2.0/tutorial/consuming_packages/cross_building_with_conan.html), I think maybe cmake-conan can generate those Profiles based on `conan_cmake_autodetect()` or other internal commands in the current workspace or the build folder.

Therefore, we can pass those profiles to the `conan_cmake_install()`. For example:

```cmake
conan_cmake_install(
  PATH_OR_REFERENCE .
  BUILD             missing
  REMOTE            conancenter
  PROFILE_HOST      ${CMAKE_BINARY_DIR}/host_profile
  PROFILE_BUILD     ${CMAKE_BINARY_DIR}/build_profile
)
```

hwhsu1231:
This PR is trying to close: https://github.com/conan-io/cmake-conan/issues/450

I implement `conan_cmake_profile()` function in `conan.cmake`, and add a test for it in `tests.py`.

The following is the test log locally.

```cmd
F:\GitRepo\cmake-conan>pytest tests.py
========================== test session starts ==========================
platform win32 -- Python 3.10.7, pytest-7.1.3, pluggy-1.0.0
rootdir: F:\GitRepo\cmake-conan
collected 38 items

tests.py ..................s...................                    [100%]

=============== 37 passed, 1 skipped in 234.98s (0:03:54) ===============
```


hwhsu1231:
The only **SHORTAGE** of the current implementation is that it **CANNOT** use `;` in `ENV`, `BUILDENV`, and `RUNENV` to generate the content like this: (See the [documentation](https://docs.conan.io/en/latest/reference/profiles.html#buildenv))

```ini
[buildenv]
# Define a variable "MyVar1"
MyVar1=My Value; other
```

It is because CMake use `;` as **the seperator of items in the list** as well. Any idea?

hwhsu1231:
I tried to follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) specification. So I modified some commit messages.

hwhsu1231:
> The only **SHORTAGE** of the current implementation is that it **CANNOT** use `;` in `ENV`, `BUILDENV`, and `RUNENV` to generate the content like this: (See the [documentation](https://docs.conan.io/en/latest/reference/profiles.html#buildenv))

Or maybe just add some notes to tell users **NOT** to use **semicolons** in those options.

czoido:
Hi @hwhsu1231,
Thanks for the contribution. I would like to understand better the need of adding this. Could you please explain a bit the use case of this function and the advantages over creating the profiles outside the CMake wrapper, having them in your project and just adding them as arguments in the `conan_cmake_install` call?

hwhsu1231:
@czoido 

This command can help some beginners to customize their Conan Profile required since they may not familiar with operations of Conan. Also, it is recommended to use **Host Profile** and **Build Profile** in Conan 2. Most importantly, I think the integration of Conan with CMake should make users **manually** operate the Conan Configuration **as less as possible**.

I already posted an example usage with `conan_cmake_install()` command right here: https://github.com/conan-io/cmake-conan/issues/450#issuecomment-1303722392

hwhsu1231:
I've rebased this branch from the latest `develop` branch.

conan.cmake
```diff
@@ -1054,3 +1055,67 @@ macro(conan_config_install)
     message(FATAL_ERROR "Conan config failed='${return_code}'")
   endif()
 endmacro()
+
+
+function(conan_cmake_profile)
+
+  conan_parse_arguments(${ARGV})
```
czoido: Instead of adding the arguments to the `conan_parse_arguments` macro, you can add the parsing inside this function. Check how the conan_cmake_install() function is implemented: 
https://github.com/conan-io/cmake-conan/blob/835529148ac0def63bbd1d07ceb4e4c130d233af/conan.cmake#L556-L560

czoido:
Hi @hwhsu1231,
I think this feature could be merged, but please check my comment above. Also, the docs are still missing, could you please add some documentation about the usage in the README?
Thanks a lot.

conan.cmake
hwhsu1231: Done. Please check.

hwhsu1231:
@czoido 

I followed what you required in the above comment and add some docs about `conan_cmake_profile()` in README.md. Please check.

conan.cmake
```diff
@@ -458,7 +458,8 @@ macro(conan_parse_arguments)
   set(oneValueArgs CONANFILE  ARCH BUILD_TYPE INSTALL_FOLDER OUTPUT_FOLDER CONAN_COMMAND)
   set(multiValueArgs DEBUG_PROFILE RELEASE_PROFILE RELWITHDEBINFO_PROFILE MINSIZEREL_PROFILE
                      PROFILE REQUIRES OPTIONS IMPORTS SETTINGS BUILD ENV GENERATORS PROFILE_AUTO
-                     INSTALL_ARGS CONFIGURATION_TYPES PROFILE_BUILD BUILD_REQUIRES)
+                     INSTALL_ARGS CONFIGURATION_TYPES PROFILE_BUILD BUILD_REQUIRES TOOL_REQUIRES 
+                     CONF BUILDENV RUNENV)
```
czoido: No need to modify this
```suggestion
                     INSTALL_ARGS CONFIGURATION_TYPES PROFILE_BUILD BUILD_REQUIRES)
```

czoido:
@hwhsu1231 just one suggestion above and after that I think we are ready to merge 🙂

conan.cmake
hwhsu1231: Done.

hwhsu1231:
@czoido

I followed what you required in the above comment. Please check.

czoido (approved):

czoido:
Merged, thanks for contributing this.

PR #458, opened 2022-11-08
Add conan_cmake_env_autodetect

uyha:
This makes it easier to cross compile by passing the appropriate environment variables to `conan_cmake_run`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,11 +46,13 @@ conan_cmake_configure(REQUIRES fmt/6.1.2
                       GENERATORS cmake_find_package)
 
 conan_cmake_autodetect(settings)
+conan_cmake_env_autodetect(env_vars)
 
 conan_cmake_install(PATH_OR_REFERENCE .
                     BUILD missing
                     REMOTE conancenter
-                    SETTINGS ${settings})
+                    SETTINGS ${settings}
+                    ${env_vars})
 
 find_package(fmt)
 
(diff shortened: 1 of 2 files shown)


czoido:
Hi @uyha,
Thanks for the contribution. Could you please give a bit more context on how this makes it easier to cross compile? Those variables should be already taken into account as they are passed in the profile... could you provide some example of the issues you are facing?
Thanks a lot

uyha:
I was trying to cross compile from a x86_64 machine to an armv7 machine. The project i was trying to cross compile has `openssl` in the dependency tree. For some reason, the build script didn't pick up the cross compiler I set through the CMAKE_C_COMPILER and CMAKE_CXX_COMPILER, and would fail to compile. This is fixed by setting the environment variables by hand (hence the PR). However, this didn't actually help me to build my project, so I just went the route of not using `conan_cmake_autodetect`, and specify the profiles for build and host machines myself.

czoido:
> I was trying to cross compile from a x86_64 machine to an armv7 machine. The project i was trying to cross compile has `openssl` in the dependency tree. For some reason, the build script didn't pick up the cross compiler I set through the CMAKE_C_COMPILER and CMAKE_CPP_COMPILER, and would fail to compile. This is fixed by setting the environment variables by hand (hence the PR). However, this didn't actually help me to build my project, so I just went the route of not using `conan_cmake_autodetect`, and specify the profiles for build and host machines myself.

Hi @uyha, specifying the profiles for build and host machines is the best way to go if you are cross-compiling, I'm glad to hear that it works now. I'm closing this as the problem seems solved. Thanks a lot for your contribution.

PR #461, opened 2022-11-18
User just major version of compiler in autodetection

czoido:
The purpouse of this is to align cmake-conan behaviour with Conan autdetected profiles. In Conan in most of the cases only the major version of the compiler is used in the autodetection. It will also not make users update their settings.yml if they update the compiler and the minor version is not still in the Conan client settings.


hwhsu1231:
@czoido 

It seems that `gcc` and `clang` compilers require `MINOR` version for some situations:

image
image

czoido:
Thanks @hwhsu1231, I wanted to keep it simple, but in the end I added some logic to consider those cases and mimic the behaviour of the Conan client: https://github.com/conan-io/conan/blob/3d927e2432d1571956946ae2cfc220c79893cc1f/conans/client/conf/detect.py#L149

memsharded (approved):

PR #471, opened 2023-01-18
Add command line flags for host profile.

TobiSchluter:
This seems to be missing. With this, I can cross-build out of the box akin to
```
conan_cmake_run(
    SETTINGS ${settings}
    ENV ${conan_env}
    PROFILE_BUILD default
    PROFILE_HOST android

    REQUIRES
    boost/1.78.0
    catch2/2.13.8
    eigen/3.4.0
    expected-lite/0.6.2
    fmt/9.1.0
    nlohmann_json/3.11.2
    openssl/1.1.1o
    valijson/0.6
    BASIC_SETUP
    CMAKE_TARGETS
    CONAN_COMMAND ${CONAN_CMD}

    OPTIONS
    boost:without_stacktrace=True
    BUILD missing

    GENERATORS cmake_find_package)
```


TobiSchluter:
I seem to have managed to mess up this patch when copying it between my trees, sorry about the two separate commits. Now it includes the whole change.

TobiSchluter:
Any interest? Just ran into this again when updating.
I have another patch which adds support for `-c` flags, but I'd like to get this out of the way first.

memsharded:
Hi @TobiSchluter 

Thanks for your contribution, and sorry for taking so long to check it. We have been focused on releasing 2.0, and also for the new ``cmake-conan`` integration based on the new CMake dependency providers, there have been a few things that have fallen through the cracks.

conan.cmake
```diff
@@ -302,6 +302,10 @@ function(conan_cmake_settings result)
         conan_check(VERSION 1.24.0 REQUIRED DETECT_QUIET)
         set(_SETTINGS ${_SETTINGS} -pr:b=${ARG})
     endforeach()
+    foreach(ARG ${ARGUMENTS_PROFILE_HOST})
+        conan_check(VERSION 1.24.0 REQUIRED DETECT_QUIET)
```
memsharded, line 306: This hardcoded check repeated for every profile seems wrong. It seems the previous one for ``PROFILE_BUILD`` was a bit legacy and dirty, it shouldn't be done that way.

TobiSchluter:
I don't remember this at all, or only barely. I'm closing it, as I cannot really do much with "it shouldn't be done this way (which is the way it is done in this file)" without digging in too deep again for a problem that is likely theoretical now, three years later. Sorry for missing this back then.

memsharded:
Thanks for following up. Yes, the paradigm has completely shifted in Conan 2 to CMake dependency providers, so this is outdated, thanks for closing!

PR #473, opened 2023-03-08
experiment with CMAKE_PROJECT_INCLUDE

memsharded:
Very preliminary
- test in ``tests_new.py`` (not run by CI)
- Windows detection only
- several TODOs in code


conaninstall.cmake
memsharded, line 1: This file can be removed

Issue #475, opened 2023-03-09
Discussion of cmake-conan evolving towards Dependency Provider.


hwhsu1231:
Recently, I found that a new experimental branch ([experiment/cmake_project_include](https://github.com/conan-io/cmake-conan/tree/experiment/cmake_project_include)) was created, and the Conan team has submitted a few PRs related to it:

* #473 
* #474
* #476

Does cmake-conan decide to evolve towards [Dependency Providers](https://cmake.org/cmake/help/latest/guide/using-dependencies/index.html#dependency-providers)?

If that is the case, then I would be very happy to see it evolves in this direction. I think this solution may be better than the cmake-conan-setup (#449) that I previously proposed.

memsharded:
It is not a decision yet, but we are experimenting with the different possibilities, trying to achieve the best possible integration.
Certainly the dependency providers looks good, but we want to learn first if there could be some blocker or problems before announcing anything, so it is expected that we will be trying things for a couple of weeks.

hwhsu1231:
@memsharded - Have you considered the CAE problem of CMakeToolchain generator in this scheme?

memsharded:
> @memsharded - Have you considered the CAE problem of CMakeToolchain generator in this scheme?

Yes, hopefully it will not be a problem

hwhsu1231:
> Yes, hopefully it will not be a problem

@memsharded 

Based on what [CMake Docs](https://cmake.org/cmake/help/latest/command/cmake_language.html#dependency-providers) said about Dependency Provider:

> When a call is made to `find_package()` or `FetchContent_MakeAvailable()`, the call may be forwarded to a dependency provider which then has the opportunity to fulfill the request.

If I haven't misunderstood the mechanism, it seems that the `xxx_provide_dependency()` function is called **_only when_** we call `find_package()` command. For example:

```cmake
cmake_minimum_required(VERSION 3.24)
project(cmake-conan-test LANGUAGES CXX)
#
# Call the xxx_provide_dependency() when the find_package(fmt) is invoked.
#
find_package(fmt)
```

That means the `conan_toolchain.cmake` file of CMakeToolchain generator will be generated _**after**_ the `project()` command. It's because the `conan install` command is executed inside the `conan_provide_dependency()` function currently.

https://github.com/conan-io/cmake-conan/blob/1f33316ad521dab334d817ca30218bb8790fa7aa/conantools.cmake#L127-L151

[...]

PR #476, opened 2023-03-13
dep provider: fixes

jcar87:
* Add logic to map CMake compiler names for gcc and clang
* Propagate the CMake generator of the current project to `tools.cmake.cmaketoolchain:generator` - this will cause the CMakeDeps generator to generate files in the same folder when calling `conan install` for both Debug and Release
* Conditionally call `conan install` twice based on `CMAKE_CONFIGURATION_TYPES` - this is only defined in multi-config generators - unfortunately `CMAKE_BUILD_TYPE` may still be defined at configure time even on a multi-config generator (even if it is a no-op)
* On non-windows platforms, test both the default generator _and_ a multi-config generator. 

The test passes on Linux and Mac now. On Linux it passes with both gcc and clang.


hwhsu1231:
> * Propagate the CMake generator of the current project to `tools.cmake.cmaketoolchain:generator` - this will cause the CMakeDeps generator to generate files in the same folder when calling `conan install` for both Debug and Release

@jcar87 - Any idea of solving the CAE problem mentioned in https://github.com/conan-io/cmake-conan/issues/475#issuecomment-1465095883?

conantools.cmake
```diff
@@ -91,6 +99,10 @@ function(detect_host_profile output_file)
     else()
         set(_FN ${output_file})
     endif()
+
+    string(APPEND PROFILE "[conf]\n")
+    string(APPEND PROFILE "tools.cmake.cmaketoolchain:generator=${CMAKE_GENERATOR}\n")
```
memsharded, line 104: Is CMAKE_GENERATOR always defined as a variable? Even when not provided in CLI and defaulted?

conantools.cmake
```diff
@@ -128,7 +140,7 @@ function(conan_provide_dependency package_name)
     if(NOT CONAN_INSTALL_SUCCESS)
         message(STATUS "CMake-conan: first find_package() found, running 'conan install' to install deps")
         detect_host_profile(${CMAKE_BINARY_DIR}/conan_host_profile)
-        if(CMAKE_BUILD_TYPE)
+        if(NOT CMAKE_CONFIGURATION_TYPES)
```
memsharded, line 143: I was relying on the fact that users in Windows can still force installing just 1 configuration with ``-DCMAKE_BUILD_TYPE=Debug``, even if that does nothing, Conan will know to install only the Debug.

If we are not allowing this, we would need another kind of opt-out I guess?

memsharded (approved):
Looking good!

conantools.cmake
jcar87: yup, should be defined by CMake itself, even when not provided by the user

conantools.cmake
jcar87: yup, we would need some opt-out indeed!  I would try to avoid giving `CMAKE_BUILD_TYPE` a meaning in the context of a multi-config generator (where giving `CMAKE_BUILD_TYPE` a value at configure time is unfortunately allowed, but will have no effect).

PR #505, opened 2023-04-21
Use presets to simplify configuration and build commands

ChrisThrasher:
CMake presets are the perfect tool to capture the particular flags required to build the `develop2` branch. This repo ought to set an example for how presets can be used to avoid typing out long configuration commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,9 +26,8 @@ This repository contains a `CMakeLists.txt` with an example project that depends
 
 ```bash
 cd cmake-conan
-mkdir build
-cmake -B build -S . -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=conan_provider.cmake -DCMAKE_BUILD_TYPE=Release
-cmake --build build --config Release
+cmake --preset default
+cmake --build --preset default
 ```
 
 ### In your own project
--- a/README.md
+++ b/README.md
@@ -38,7 +38,6 @@ cmake --build --preset default
 
 ```bash
 cd [your-project]
-mkdir build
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -25,10 +25,8 @@ git clone https://github.com/conan-io/cmake-conan.git -b develop2
 This repository contains a `CMakeLists.txt` with an example project that depends on `fmt`. 
 
 ```bash
-cd cmake-conan
```
memsharded, line 28: The ``cd cmake-conan`` is still necessary, we are not yet inside the cloned project

memsharded (commented):
thanks for your contribution!

README.md
```diff
@@ -26,9 +26,8 @@ This repository contains a `CMakeLists.txt` with an example project that depends
 
 ```bash
 cd cmake-conan
-mkdir build
-cmake -B build -S . -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=conan_provider.cmake -DCMAKE_BUILD_TYPE=Release
-cmake --build build --config Release
+cmake --preset default
```
memsharded, line 29: Let's leave both approaches.
I think it is good that we provide both, with and without presets, for users that don't want to use presets. Could you please leave it?

memsharded:
Hi @ChrisThrasher 
Did you accidentally close the PR? Sorry for the delay, it didn't mean that it was not an interesting contribution, but mostly that we didn't have the time to manage it.

I would still like @jcar87 feedback if he feels it would be good to add these presets.

ChrisThrasher:
I closed it because I lost motivation to complete it and didn't want lingering open PRs around.

memsharded:
Ok, no prob, we might take the idea of adding these presets and add it ourselves. Thanks anyway!

PR #506, opened 2023-04-27
feat: reconfigure on conanfile changes

tim-goto:
This will reconfigure on changes in the conanfile and also move the conan install success to a global property instead of a cache variable.

### DIFF
--- a/conan_support.cmake
+++ b/conan_support.cmake
@@ -147,18 +147,24 @@ function(conan_install)
         message(FATAL_ERROR "Conan install failed='${return_code}'")
     else()
         # the files are generated in a folder that depends on the layout used, if
-        # one if specified, but we don't know a priori where this is. 
+        # one is specified, but we don't know a priori where this is.
         # TODO: this can be made more robust if Conan can provide this in the json output
         string(JSON CONAN_GENERATORS_FOLDER GET ${conan_stdout} graph nodes 0 generators_folder)
         # message("conan stdout: ${conan_stdout}")
         message(STATUS "CMake-Conan: CONAN_GENERATORS_FOLDER=${CONAN_GENERATORS_FOLDER}")
-        set(CONAN_GENERATORS_FOLDER "${CONAN_GENERATORS_FOLDER}" PARENT_SCOPE)
-        set(CONAN_INSTALL_SUCCESS TRUE CACHE BOOL "Conan install has been invoked and was successful")
+        set_property(GLOBAL PROPERTY CONAN_GENERATORS_FOLDER "${CONAN_GENERATORS_FOLDER}")
+        # reconfigure on conanfile changess
+        string(JSON CONANFILE GET ${conan_stdout} graph nodes 0 label)
+        message(STATUS "CMake-Conan: CONANFILE=${CMAKE_SOURCE_DIR}/${CONANFILE}")
+        set_property(DIRECTORY ${CMAKE_SOURCE_DIR} APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS "${CMAKE_SOURCE_DIR}/${CONANFILE}")
+        # success
+        set_property(GLOBAL PROPERTY CONAN_INSTALL_SUCCESS TRUE)
     endif()
 endfunction()
 
 
 macro(conan_provide_dependency package_name)
+    get_property(CONAN_INSTALL_SUCCESS GLOBAL PROPERTY CONAN_INSTALL_SUCCESS)
     if(NOT CONAN_INSTALL_SUCCESS)
         find_program(CONAN_COMMAND "conan" REQUIRED)
         message(STATUS "CMake-Conan: first find_package() found. Installing dependencies with Conan")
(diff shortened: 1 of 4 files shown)


conan_support.cmake
```diff
@@ -1,6 +1,6 @@
 function(detect_os OS)
     # it could be cross compilation
-    message(STATUS "Conan-cmake: cmake_system_name=${CMAKE_SYSTEM_NAME}")
+    message(STATUS "CMake-Conan: cmake_system_name=${CMAKE_SYSTEM_NAME}")
```
memsharded: It is typically better to not modify this kind of thing in PRs like this one. We can definitely change it if we think it is better this way, but better in another PR, and that helps keeping the PRs as short as possible, which helps reviewers too.

conan_support.cmake
```diff
@@ -147,39 +147,47 @@ function(conan_install)
...
         # the files are generated in a folder that depends on the layout used, if
-        # one if specified, but we don't know a priori where this is. 
+        # one is specified, but we don't know a priori where this is.
         # TODO: this can be made more robust if Conan can provide this in the json output
         string(JSON CONAN_GENERATORS_FOLDER GET ${conan_stdout} graph nodes 0 generators_folder)
         # message("conan stdout: ${conan_stdout}")
-        message(STATUS "CMake-conan: CONAN_GENERATORS_FOLDER=${CONAN_GENERATORS_FOLDER}")
-        set(CONAN_GENERATORS_FOLDER "${CONAN_GENERATORS_FOLDER}" PARENT_SCOPE)
-        set(CONAN_INSTALL_SUCCESS TRUE CACHE BOOL "Conan install has been invoked and was successful")
+        message("extracted: ${CONAN_GENERATORS_FOLDER}")
```
memsharded: This will print twice the same folder, probably not necessary.

conan_support.cmake
```diff
@@ -147,39 +147,47 @@ function(conan_install)
...
+        message(STATUS "CMake-Conan: CONAN_GENERATORS_FOLDER=${CONAN_GENERATORS_FOLDER}")
+        set_property(GLOBAL PROPERTY CONAN_GENERATORS_FOLDER "${CONAN_GENERATORS_FOLDER}")
+        # reconfigure on conanfile changess
+        string(JSON CONANFILE GET ${conan_stdout} graph nodes 0 label)
+        message(STATUS "CMake-Conan: CONANFILE=${CMAKE_SOURCE_DIR}/${CONANFILE}")
+        set_property(DIRECTORY ${CMAKE_SOURCE_DIR} APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS "${CMAKE_SOURCE_DIR}/${CONANFILE}")
+        # success
+        set_property(GLOBAL PROPERTY CONAN_INSTALL_SUCCESS TRUE)
```
memsharded, line 161: Could you please elaborate a bit more this change from Cache variable to global property? What are the advantages/differences in behavior?

memsharded (commented):
This looks like a good addition, thanks very much for your contribution!

conan_support.cmake
tim-goto: Sorry, I created #508 for that

tim-goto: I will change this PR in a second

conan_support.cmake
tim-goto: The cache variable would prevent a reconfiguration when the conanfile changed since it will still be set from a previous configuration run

conan_support.cmake
tim-goto: Sorry, forgot to remove that

tim-goto:
I will rebase and fix CI in a sec

memsharded:
> I will rebase and fix CI in a sec

Thanks. I am experimenting and testing a little bit your branch.

Why the rename of the ``tests.py`` -> ``test_smoke.py``? This seems unnecessary, when we have more tests, we will create a ``tests`` folder and have there different named files.

tim-goto:
I just checked the pytest guidelines and they recommend doing it so, see [here](https://docs.pytest.org/en/7.3.x/how-to/usage.html)

> In general, pytest is invoked with the command pytest (see below for [other ways to invoke pytest](https://docs.pytest.org/en/7.3.x/how-to/usage.html#invoke-other)). This will execute all tests in all files whose names follow the form test_*.py or \*_test.py in the current directory and its subdirectories. More generally, pytest follows [standard test discovery rules](https://docs.pytest.org/en/7.3.x/explanation/goodpractices.html#test-discovery).

Please let me know if I should change that back.

memsharded:
> Please let me know if I should change that back.

Ah, I see. It is not a problem in general, it works as long as ``test`` is in the name, you can leave it as-is

memsharded (approved):
I have been testing this and overall it seems a good improvement, many thanks! 
So I am going to merge it.
The calls to ``conan install`` at ``cmake`` build stage when something changes is good.

There is however a change in behavior when cmake configure step is repeatedly called. For every invocation it will run ``conan install`` again, even if there is no change to ``conanfile.txt``. While this seems a safer default to account for changes, it is true that there could be other problems, like the CMake cache being invalid and other issues. So we might investigate this and change it, so it only re-runs ``conan install`` when the conanfile changes.

PR #507, opened 2023-04-27
feat: detect conan command with paths suggestions

tim-goto:
This allows setting the path to the conan executable through cmake variables in the calling `CMakeLists.txt`

### DIFF
--- a/conan_support.cmake
+++ b/conan_support.cmake
@@ -109,9 +109,19 @@ function(detect_host_profile output_file)
 endfunction()
 
 
+function(detect_conan_command)
+    if(DEFINED CONAN_COMMAND_PATHS)
+        find_program(CONAN_COMMAND "conan" PATHS ${CONAN_COMMAND_PATHS} REQUIRED NO_CMAKE_FIND_ROOT_PATH NO_DEFAULT_PATH)
+    else()
+        find_program(CONAN_COMMAND "conan")
+    endif()
+    set(CONAN_COMMAND ${CONAN_COMMAND} CACHE FILEPATH "Path to the Conan executable")
+endfunction()
+
+
 function(conan_profile_detect_default)
     message(STATUS "Conan-cmake: Checking if a default profile exists")
-    execute_process(COMMAND conan profile path default
+    execute_process(COMMAND ${CONAN_COMMAND} profile path default
                     RESULT_VARIABLE return_code
                     OUTPUT_VARIABLE conan_stdout
                     ERROR_VARIABLE conan_stderr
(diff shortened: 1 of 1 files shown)


conan_support.cmake
```diff
@@ -109,9 +109,19 @@ function(detect_host_profile output_file)
 endfunction()
 
 
+function(detect_conan_command)
+    if(DEFINED CONAN_COMMAND_PATHS)
+        find_program(CONAN_COMMAND "conan" PATHS ${CONAN_COMMAND_PATHS} REQUIRED NO_CMAKE_FIND_ROOT_PATH NO_DEFAULT_PATH)
+    else()
+        find_program(CONAN_COMMAND "conan")
```
memsharded: Is there any possibility that this will find a Conan different to the one that command line invocation would use? Like when being inside a virtualenv or something like that?

memsharded (commented):
I am not fully sure about this. 
One of the greatest things of the new CMake integrations and this proposed dependency-provider approach is that it doesn't need to add Conan things to the ``CMakeLists.txt``.

Having to add a ``set(CONAN_COMMAND_PATHS /some/hardcoded/path)`` looks a bit like an anti pattern, needing to define things in a CMakeLists that are both Conan-specific and also hardly portable to different computers, platforms, etc. Elaborating a bit more on the use case would be good.

conan_support.cmake
tim-goto: If you provide paths suggestions it will find the conan that is present in those. E.g. if you install conan with apt and in an virtualenv you could not tell for sure which one will be found first without looking at the path. With this you can provide suggestions as to which one you want to use.

tim-goto:
> I am not fully sure about this. One of the greatest things of the new CMake integrations and this proposed dependency-provider approach is that it doesn't need to add Conan things to the `CMakeLists.txt`.
> 
> Having to add a `set(CONAN_COMMAND_PATHS /some/hardcoded/path)` looks a bit like an anti pattern, needing to define things in a CMakeLists that are both Conan-specific and also hardly portable to different computers, platforms, etc. Elaborating a bit more on the use case would be good.

I can understand your concerns, the current approach is very clean. Our use-case is the following: instead of using the system package manager or pip/... to install conan, we rely on an entire infrastructure of CMake scripts to package and install build dependencies. This means, that we use some CMake code to install conan to a tools directory that is not in path. In order to provide the `conan_support.cmake` with the path to our conan installation, we need to set something in our `CMakeLists.txt` anyways. How about instead of a 'magic variable' I create a function along the lines `conan_cmake_setup()` that is well documented and can accept multiple parameters to allow customisation of the underlying conan commands? For example setting the conan executable path but also provide a way to customize the detected host profile? Please let me know what you think!

tim-goto:
What do you think of this approach? I am still not sure how to implement a function that will allow the user to set parameters of the conan command line calls, do you have an opinion on this? I think it would also be useful for other people.

memsharded (approved):
Let's merge this, this wouldn't need an extra CONAN_XXX variable that would pollute the user ``CMakeLists.txt``, but can rely on standard CMake ``find_program()`` paths)

PR #510, opened 2023-05-03
feat: add conan_configure to set install params

tim-goto:
@memsharded please let me know what you think about this. It would allow a user to configure the parameters passed to the conan install command. If you think this is useful, I will write a test as well.


conan_support.cmake
```diff
@@ -132,12 +132,9 @@ endfunction()
 
 
 function(conan_install)
-    cmake_parse_arguments(ARGS CONAN_ARGS ${ARGN})
-    set(CONAN_OUTPUT_FOLDER ${CMAKE_BINARY_DIR}/conan)
     # Invoke "conan install" with the provided arguments
-    set(CONAN_ARGS ${CONAN_ARGS} -of=${CONAN_OUTPUT_FOLDER})
-    message(STATUS "CMake-conan: conan install ${CMAKE_SOURCE_DIR} ${CONAN_ARGS} ${ARGN}")
-    execute_process(COMMAND conan install ${CMAKE_SOURCE_DIR} ${CONAN_ARGS} ${ARGN} --format=json
+    message(STATUS "CMake-conan: conan install ${CMAKE_SOURCE_DIR} ${ARGN}")
```
memsharded: This change looks good, having ``conan_install`` to not have any custom logic, just execute the args.

conan_support.cmake
```diff
@@ -183,3 +186,91 @@ macro(conan_provide_dependency package_name)
     endif()
     find_package(${ARGN} BYPASS_PROVIDER)
 endmacro()
+
+
+function(conan_configure)
```
memsharded: It is not very clear the purpose of this generalization, with management of all arguments, but that might not be necessary most of them?

memsharded (commented):
The use-case or problem that this is trying to solve is not evident. It looks like a cleaner structure overall, but not very clear why all arguments need to be managed. If you can elaborate a bit more about the intended usage of this on the user side, it would help. Thanks!

tim-goto:
Thanks for your feedback. Currently, it is not possible to influence any of the parameters of the conan install at all. This would provide a way for a user to e.g. set fields in the host/build profile from CMake during the configure phase (our intended use-case). Not all parameters might be required (I also left out the `-h` flag because it doesn't make sense) but it seems like this offers the most flexibility. If you have a preference which parameters to support, please let me know and I will update the PR accordingly.

memsharded:
> Thanks for your feedback. Currently, it is not possible to influence any of the parameters of the conan install at all. This would provide a way for a user to e.g. set fields in the host/build profile from CMake during the configure phase (our intended use-case). Not all parameters might be required (I also left out the -h flag because it doesn't make sense) but it seems like this offers the most flexibility. If you have a preference which parameters to support, please let me know and I will update the PR accordingly.

I think implementing the full mapping/translation of all ``conan install`` arguments to CMake arguments is too much. Probably a way to define arguments as raw strings already in the ``conan install`` cli format would be enough.

We are also a bit concerned about the mismatch between the Conan inputs and the current build. Once you define ``compiler=clang`` for example in your Conan profile, many users will expect ``CMake`` to use ``clang`` in the current build. This is not possible, because the processing of this happens later, at ``find_package()`` time, when the CMake toolchain has already been defined, lets say to use ``gcc`` instead. Then, all sorts of binaries incompatibilities can happen, and users will be blaming Conan and the integration for the mismatch.

We still envision the possible flows as follow:
- If you want CMake to be the main driver, then the configuration defined by CMake is the one that matters, and it doesn't make sense to have users changing the configuration to call ``conan install`` to a different one.
- If you want the Conan configuration to be the main driver, then the flow is ``conan install`` + ``cmake --presets conan-release``, and you can leverage the generated ``conan_toolchain.cmake`` and the generated presets to account for the configuration.

What seems complicated or problematic is the hybrid, you want to use configuration from Conan (profiles, settings, options), but don't want to call Conan first, but call CMake only. Then we have the chicken-and-egg problem that doesn't have a good solution, and that can cause troubles to many users. It is obvious that you know what you are doing and you will probably not have this problem if this is merged, but I am quite sure that we will get many reports from other users because of issues derived from this.

tim-goto:
Thanks for your feedback, I will discuss internally and get back to you/close the PR within the next couple of weeks.

tim-goto:
Hey @memsharded , I rebased the PR and tried to simplify a bit. I still think it is a valid use-case for consumers to pass parameters to the conan install command. For example, if I want to use a custom remote, that is currently not possible. The function would allow setting all parameters and just adds a convenience wrapper for the parameters that can handle multiple inputs. When #515 is merged I will also be able to add some tests for this. As for the possible foot-guns, maybe a hint in the readme would suffice: If you are using the `conan_configure()` to tweak the conan install command and encounter errors, please resolve them yourself.

tim-goto:
Hey @memsharded , I added some tests, please let me know what you think.

tim-goto:
Hey @memsharded , please let me know what you think of this PR

jcar87:
Hi @tim-goto  - we have merged the ability to customize the profile:
https://github.com/conan-io/cmake-conan/tree/develop2#customizing-conan-profiles

I believe that with the ability to customize the profile and also having control of `conanfile.txt`, this should cover all options/arguments you are proposing, with the exception of `--build` (which for the time being is still set to `--build=missing`).

We are opting to minimise the interface to interact with Conan via CMake as much as we possibly can, for the sake of simplicity. Otherwise we will get feature creep of trying to replicate in CMake variables how to control the Conan CLI, and down the line that will be difficult to maintain.

PR #515, opened 2023-05-10
Extend testing

tim-goto:
Please let me know if there are too many changes in this PR. I tried to split the tests according to the pytest [anatomy of a test](https://docs.pytest.org/en/7.1.x/explanation/anatomy.html). Also this should be merged after #514 because it already contains that commit.


conanfile.txt
```diff
@@ -1,5 +0,0 @@
-[requires]
```
memsharded: I think we don't want to remove this, as this is referenced in the README, as the getting started.
Probably we want to put it inside an ``examples`` folder?

resources/basic/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 3.24)
```
memsharded: with these changes, I think it is time to put all of this under a ``tests`` folder of the project.

memsharded (commented):
This looks great, thanks very much for your contribution!

conanfile.txt
tim-goto: Sure, I can restore the example files and put them in a separate folder. At the moment I just moved the reference in the README to the test fixtures folder.

memsharded: Yes, I think the readme instructions with an example that users can reproduce to test things is important, so I'd prefer to keep it. I'd say to leave these 2 files in place for this PR, and it can be improved in a separate PR, let this one be focused on the testing.

memsharded (approved):

PR #516, opened 2023-05-17
feat: configure runtime settings for MSVC based on CMAKE_MSVC_RUNTIME_LIBRARY

mkviatkovskii:
This pull request allows to set `compiler.runtime` and `compiler.runtime_type` settings based on `CMAKE_MSVC_RUNTIME_LIBRARY`.


memsharded (commented):
Looking good, thanks very much for your contribution!

memsharded (approved):

conan_support.cmake
```diff
@@ -39,6 +39,22 @@ function(detect_compiler COMPILER COMPILER_VERSION)
     if(_COMPILER MATCHES MSVC)
         set(_COMPILER "msvc")
         string(SUBSTRING ${MSVC_VERSION} 0 3 _COMPILER_VERSION)
+        # Configure compiler.runtime and compiler.runtime_type settings for MSVC
+        if (CMAKE_MSVC_RUNTIME_LIBRARY)
```
jcar87: Perhaps for consistency I'd consider a `DEFINED` here, as in line 25? Although it does not make a practical difference in this case.

On the other hand, I believe this conditional does not cover the most common use-case, that is, leaving CMake to pick-up the default "Visual Studio" generator, which is multi-configuration. For such a generator, I'd expect that if `CMAKE_MSVC_RUNTIME_LIBRARY` is defined, it is conditional on the build type, e.g.  `MultiThreaded$<$<CONFIG:Debug>:Debug>`  (as per documentation [here](https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html)) - and because of the use of `STREQUAL` this is not covered here. 


If a project specified `MultiThreaded$<$<CONFIG:Debug>:Debug>` (the `static` compiler runtime) - it would not be covered by this conditional, we would then only pass `compiler.build_type` as per https://github.com/conan-io/cmake-conan/blob/a261d04f3ce117ec31accb0795b93c7f8a86e452/conan_support.cmake#L207C19-L209, which may fall back to the default `dynamic` runtime, causing a mismatch.

On the other hand, if a project using a multi-config generator like Visual Studio specifies a single fixed value (e.g. `MultiThreadedDebugDLL`), that would be used also for the release configuration, and I'd imagine there may be binary incompatibilities.

Perhaps a way to make this more robust - we need to consider two axes:
* whether `CMAKE_MSVC_RUNTIME_LIBRARY` is defined, and if it is, whether it contains a generator expression (`$<`) 
* whether we are in a single or multi configuration generator (testing the presence of `CMAKE_CONFIGURATION_TYPES`)


If we are in a single configuration generator:
      * and `CMAKE_MSVC_RUNTIME_LIBRARY` is defined and matches one of the four values in this list (no generator expression), then it makes sense to define `compiler.runtime` and `compiler.runtime_type` with this exact mapping)
      * if `CMAKE_MSVC_RUNTIME_LIBRARY` is defined and contains a generator expression (we could test for the specific $<CONFIG: string) - then we need to test for the presence of `DLL` in the string to define `compiler.runtime`, and let Conan derive the `runtime_type` based on the `build_type` setting (as per https://github.com/conan-io/cmake-conan/blob/a261d04f3ce117ec31accb0795b93c7f8a86e452/conan_support.cmake#L88-L92)
      
If we are in a multi configuration generator:
    * If `CMAKE_MSVC_RUNTIME_LIBRARY` is one of these four values with no generator expression, this is likely an error - it's almost certain one of the configurations will not build at all due to incompatibilities
    * If `CMAKE_MSVC_RUNTIME_LIBRARY` is defined and contains a generator expression on the build type, same as before: evaluate the presence of `DLL` to determine the `runtime_type`, and let the `build_type` derive the rest (as per https://github.com/conan-io/cmake-conan/blob/a261d04f3ce117ec31accb0795b93c7f8a86e452/conan_support.cmake#L207-L209)

jcar87 (commented):
This is a very good contribution, thanks @mkviatkovskii.

I've left a comment as I believe this may cause issues when using multi-config generators in CMake

conan_support.cmake
mkviatkovskii: Thank you @jcar87 for detailed response! I totally agree, going to try your approach tomorrow.

memsharded (approved):

jcar87:
@mkviatkovskii, @memsharded 
I've updated the PR branch and rebased it on top of the most recent `develop2`, as well as implemented the discussed changes to support multi-configuration generators, e.g. `MultiThreaded$<$<CONFIG:Debug>:Debug>`, which I expect to be the more common case.

Additionally, have expanded the test to cover this setting to ensure that there are no compiler or linker errors due to mismatches, and that the entire dependency graph is built with the same msvc runtime as specified by `CMAKE_MSVC_RUNTIME_LIBRARY`. We've added some diagnostics to the `cmake_lib` template generated by `conan`, such that it also prints the msvc runtime the library was built with - this helps with the tests.

Thanks a lot for your contribution, as this is a much needed feature.

Please be advised that you need to review the CLA https://cla-assistant.io/conan-io/cmake-conan?pullRequest=516 again!

memsharded (approved):

memsharded:
Looks good!
Please @mkviatkovskii it seems it would be necessary to sign the CLA again, can you please have a look?

jcar87 (approved):

jcar87:
Hi @mkviatkovskii  - we are still waiting for you to review and sign the updated Contributor License Agreement before being able to merge this PR, thanks!
https://cla-assistant.io/conan-io/cmake-conan?pullRequest=516

jcar87:
Hi @mkviatkovskii  - please see message above.

dvirtz:
I need this feature but it looks like @mkviatkovskii is not responding since opening this PR. 
Is there a way to have this merged? Maybe in a separate PR?

.github/workflows/cmake_conan.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         with:
           python-version: "3.10"
       - name: Install Conan
-        run: pip install conan pytest && conan --version
+        run: pip install git+https://github.com/conan-io/conan.git@release/2.0 pytest && conan --version
```
memsharded: This shouldn't be necessary anymore
```suggestion
        run: pip install conan pytest && conan --version
```

.github/workflows/cmake_conan.yml
```diff
@@ -38,7 +38,7 @@ jobs:
         with:
           python-version: "3.10"
       - name: Install Conan
-        run: pip install conan && conan --version
+        run: pip install git+https://github.com/conan-io/conan.git@release/2.0 && conan --version
```
memsharded:
```suggestion
        run: pip install conan && conan --version
```

conan_provider.cmake
```diff
@@ -77,6 +77,39 @@ function(detect_compiler COMPILER COMPILER_VERSION)
     if(_COMPILER MATCHES MSVC)
         set(_COMPILER "msvc")
         string(SUBSTRING ${MSVC_VERSION} 0 3 _COMPILER_VERSION)
+        # Configure compiler.runtime and compiler.runtime_type settings for MSVC
+        if(CMAKE_MSVC_RUNTIME_LIBRARY)
```
memsharded, line 120: Lets move this to its own function, please

memsharded: Not necessary in this PR, I can do this later.

mkviatkovskii:
> I need this feature but it looks like @mkviatkovskii is not responding since opening this PR. Is there a way to have this merged? Maybe in a separate PR?

Sorry for absence, I've signed the agreement. @memsharded, could you please let me know if there is anything else I should do to merge this pull request, or it's already non required?

memsharded:
> Sorry for absence, I've signed the agreement. @memsharded, could you please let me know if there is anything else I should do to merge this pull request, or it's already non required?

No problem, thanks very much for signing it, and for your contribution!
Just merged it, lets move it from there. 

@dvirtz you got it, feedback welcome!

dvirtz:
Thanks @memsharded and @mkviatkovskii, this works fine for me

Issue #512, opened 2023-05-09
[develop2] General feedback


ssrobins:
Thanks for supporting this very important workflow in Conan 2.0!

I used the original CMake wrapper in [personal projects](https://github.com/ssrobins/sdl2-example/commit/2730b8cc1d08ba282f4bc946f16b3a93c07d566b) and I still use it at work. I only stopped using it so I could use Conan 2.0 beta, then I converted to [something like this](https://github.com/ssrobins/sdl2-example/commit/96108702552dc6b3d1de6d86d6a9fdbfe8765298).

...and now I'm trying `develop2` out!

Here are the main reasons why I use the CMake wrapper:
- I want exactly one interface to the build. For Conan recipes, that's Conan, but for apps built in CMake that use Conan packages, I prefer using CMake directly
- I always want the Conan package settings to match the CMake build settings as much as possible and not rely on external settings such as profiles, which may be incomplete or not match the CMake build
- I want every build configuration set in multi-config CMake generators to always have a Conan package automatically
- I want to keep it possible to use the CMake GUI and Visual Studio's 'Open Folder' functionalities which, last I checked, use CMake as its sole interface

Onto what I like about `develop2`:
- It's flexible, I can use it at the command line, I can include it into my CMake directly, or I can even call the individual functions to inject alternate settings such as alternate `--build` flag or add `--update` into `conan install`.
- The individual `detect` functions are an excellent abstraction. I even started adding additional settings to support iOS and Android and it was going together very smoothly! (happy to share)
- Generating the profiles is a great idea. When Conan 2.0 forced me to use profiles, I added the bare minimum to the files and then added the rest as command line parameters to `conan install`. Generating the whole file and then referencing it in `conan install` is way cleaner with fewer files.

Here's what I'd love for you to please consider changing (and with your blessing, I'm happy to contribute all of these changes):
- Add a check for the minimum supported version of Conan in the CMake wrapper and make it a one-line change to update it in the future. This would help new users get started faster. Right now, the errors you get with an unsupported Conan don't make it immediately apparent the Conan version is to blame. https://github.com/conan-io/cmake-conan/pull/517
- Combine `conan_provider.cmake` and `conan_support.cmake` into one file so there's only one file to copy or download into my codebase. https://github.com/conan-io/cmake-conan/pull/519
[...]

PR #517, opened 2023-05-18
Check required Conan version

ssrobins:
This adds a Conan version check so people don't get stuck trying to get this to work on an unsupported Conan version. Recently discussed in https://github.com/conan-io/cmake-conan/issues/512.

In addition to adding this functionality, I set up a rudimentary cmake unit test layer to validate Conan version parsing and version check scenarios. I see there's already some extensive test changes going on in https://github.com/conan-io/cmake-conan/pull/515, but I decided not to base this branch on that because it would make the diff harder to read. I did locate the tests in a `tests` subdir to try to follow the pattern from that pull request.

Look at the [previous checks](https://github.com/conan-io/cmake-conan/pull/517/checks) to see temporary breaks in the minimum version and the unit tests to prove they work and fail GitHub Actions.

Anyway, this is my first contribution to the repo so I'm still learning. As always, I'm open to changes to make it better and conform to the codebase!


.github/workflows/cmake_conan.yml
```diff
@@ -22,4 +22,6 @@ jobs:
         with:
           cmake-version: '3.25.x'
       - name: Run Tests
-        run: pytest -rA -v
+        run: |
+          cmake -S tests -B build_unittests
```
memsharded: I think it would be better to keep all tests under the ``pytest`` framework.
It could be a test that is just a line that ``run("cmake ....")``, but also better to run them locally, etc.

conan_support.cmake
```diff
@@ -163,10 +166,104 @@ function(conan_install)
...
+    list(GET CONAN_MINIMUM_VERSION_LIST 2 CONAN_MINIMUM_VERSION_PATCH)
+
+    conan_version_parse(parse_result CONAN_VERSION_CHECK_CURRENT ${CONAN_VERSION_CHECK_CURRENT})
+    string(REPLACE "." ";" CONAN_CURRENT_VERSION_LIST "${CONAN_VERSION_CHECK_CURRENT}")
+    list(LENGTH CONAN_CURRENT_VERSION_LIST CONAN_CURRENT_VERSION_LIST_LENGTH)
+    list(GET CONAN_CURRENT_VERSION_LIST 0 CONAN_CURRENT_VERSION_MAJOR)
+    list(GET CONAN_CURRENT_VERSION_LIST 1 CONAN_CURRENT_VERSION_MINOR)
+    list(GET CONAN_CURRENT_VERSION_LIST 2 CONAN_CURRENT_VERSION_PATCH)
+
+    if(NOT CONAN_CURRENT_VERSION_MAJOR EQUAL CONAN_MINIMUM_VERSION_MAJOR)
```
memsharded: I am not sure why not using CMake version comparison utilities ``VERSION_LESS_EQUAL``

memsharded (commented):
Thanks for your contributioN!

.github/workflows/cmake_conan.yml
ssrobins: Sure, I can do that, any preference on adding it to `test_smoke.py` versus creating a new `test_unittest.py`?

conan_support.cmake
ssrobins: It wasn't intentional, I wasn't are of that CMake feature, though I did look for it! It's fixed now.

.github/workflows/cmake_conan.yml
memsharded: A new ``test_conan_version.py`` sounds good.

conan_support.cmake
```diff
@@ -163,10 +166,78 @@ function(conan_install)
 endfunction()
 
 
+function(conan_version_parse result conan_version conan_version_raw)
```
memsharded: I think I would make it an internal implementation detail of ``conan_get_current_version``, to avoid cluttering ``conan_support.cmake`` with many public-like functions

conan_support.cmake
```diff
@@ -163,10 +166,78 @@ function(conan_install)
...
+    string(REPLACE "." ";" conan_version_list "${_conan_version}")
+    list(LENGTH conan_version_list conan_version_list_length)
+    if(conan_version_list_length EQUAL 3)
+        set(${result} TRUE PARENT_SCOPE)
+    endif()
+    set(${conan_version} ${_conan_version} PARENT_SCOPE)
+endfunction()
+
+
+function(conan_get_current_version conan_command conan_current_version)
```
memsharded: ``conan_get_version`` sounds shorter but still very meaningful, maybe the ``current`` can be dropped

conan_support.cmake
```diff
@@ -163,10 +166,78 @@ function(conan_install)
...
+    endif()
+        if(NOT CONAN_VERSION_CHECK_CURRENT)
+        message(FATAL_ERROR "CMake-Conan: Required parameter CURRENT not set!")
+    endif()
+
+    conan_version_parse(parse_result CONAN_VERSION_CHECK_MINIMUM ${CONAN_VERSION_CHECK_MINIMUM})
+    conan_version_parse(parse_result CONAN_VERSION_CHECK_CURRENT ${CONAN_VERSION_CHECK_CURRENT})
+
+    if(CONAN_VERSION_CHECK_CURRENT VERSION_GREATER_EQUAL CONAN_VERSION_CHECK_MINIMUM)
+        set(${result} TRUE PARENT_SCOPE)
```
memsharded: Maybe if we ``FATAL_ERROR`` in this point, the usage of the check function is simpler?

conan_support.cmake
```diff
@@ -163,10 +166,78 @@ function(conan_install)
...
+    set(${result} FALSE PARENT_SCOPE)
+
+    if(NOT CONAN_VERSION_CHECK_MINIMUM)
+        message(FATAL_ERROR "CMake-Conan: Required parameter MINIMUM not set!")
+    endif()
+        if(NOT CONAN_VERSION_CHECK_CURRENT)
+        message(FATAL_ERROR "CMake-Conan: Required parameter CURRENT not set!")
+    endif()
+
+    conan_version_parse(parse_result CONAN_VERSION_CHECK_MINIMUM ${CONAN_VERSION_CHECK_MINIMUM})
```
memsharded: Not very clear why parsing the version is necessary. I think that CMake knows how to do comparison of version, no need to do an actual list of integers?

conan_support.cmake
memsharded: In any case, the purpose of this function is not fully clear. CMake knows how to compare versions as strings, no need for manual parsing.

memsharded (commented):
A few more comments, thanks!

conan_support.cmake
ssrobins: Makes sense, renamed!

conan_support.cmake
ssrobins: Yeah, I was using `result` to make the function more testable, but with nothing left to test now that I'm using CMake's built-in version comparison, I moved `FATAL_ERROR` inside the function and removed the unit tests.

conan_support.cmake
ssrobins: Yeah, it doesn't really serve a purpose now that the code uses CMake's version comparison so it's now deleted.

ssrobins:
Thanks for the feedback, I think it resulted in great improvements. Let me know if you see anything else!

memsharded (approved):
Looks good, thanks very much

memsharded (approved):

memsharded:
Merged, thanks for your contribution!

ssrobins:
> Merged, thanks for your contribution!

Thanks!! I can squash things down a bit next time, wasn't sure if that was going to be automatic or not.

memsharded:
> Thanks!! I can squash things down a bit next time, wasn't sure if that was going to be automatic or not.

Not necessary, it was my bad. I usually have the "Squash & merge" Github button activated, but this time I didn't realize it was just merging without squash.

PR #518, opened 2023-05-19
Detect os.version from CMAKE_OSX_DEPLOYMENT_TARGET

ssrobins:
When [CMAKE_OSX_DEPLOYMENT_TARGET](https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_DEPLOYMENT_TARGET.html) is set on Apple platforms, use it to set the `os.version` Conan setting so all Conan packages are built with the same minimum OS version flag as the CMake targets.


conan_support.cmake
```diff
@@ -13,6 +13,14 @@ function(detect_os OS)
 endfunction()
 
 
+function(detect_os_version OS_VERSION)
```
memsharded: If this is coupled to the detection of the OS, I would make it part of ``detect_os``, and make that return both MYOS and MYOS_VERSION

memsharded (commented):
I see that the CI is failing for OSX, which looked a bit suspicious after a OSX related change, but I don't see a direct correlation with it, so lets see how next CI build goes, in case it was a glitch.

ssrobins:
> I see that the CI is failing for OSX, which looked a bit suspicious after a OSX related change, but I don't see a direct correlation with it, so lets see how next CI build goes, in case it was a glitch.

Unfortunately, it's happened multiple times, yet I can't locally reproduce it. Not sure what's going on. The runner version `2.304.0` is the same as the one used in my other pull request so I don't think it's a runner change. I'll address the other feedback and troubleshoot this more later today....

memsharded:
Yes, it is weird, seems more CI related, maybe the build is broken, we will also have a look when possible.

conan_support.cmake
ssrobins: Done

ssrobins:
> Yes, it is weird, seems more CI related, maybe the build is broken, we will also have a look when possible.

I found the problem. For some reason, the CI environment has CMake variable `CMAKE_OSX_DEPLOYMENT_TARGET` set to 12.6. When that value is used for the `os.version` Conan setting, it won't work because that value isn't in the default settings.yml. I still don't know why `CMAKE_OSX_DEPLOYMENT_TARGET` is set, it's usually blank by default.

I also don't know why the pytest output isn't showing the real error, which made it really hard to figure out. Instead, it shows the `ERROR: Profile not found: default`, which doesn't fail the build.

Anyway, I'm currently working around the problem by setting `CMAKE_OSX_DEPLOYMENT_TARGET` to a value in settings.yml.

test_smoke.py
```diff
@@ -61,6 +61,9 @@ def test1():
         cmake_minimum_required(VERSION 3.24)
         project(MyApp CXX)
 
+        if(${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")
```
memsharded: This is a bit concerning, if we don't define this in the ``CMakeLists.txt``, would it fail?
If it doesn't fail, great, then it would be worth to add a different, new test, that defined it, and then try to check that the value is actually being used.

ssrobins: In my local environment, `CMAKE_OSX_DEPLOYMENT_TARGET` is not set so `os.version` doesn't get set and everything is fine. I'm still not sure why the CI environment behaves differently, but I'll keep working on it.

After that's fixed, then yes, I plan to add test coverage for the situation that `CMAKE_OSX_DEPLOYMENT_TARGET` is set to a value in the default settings.yml file as of the minimum Conan version supported.

tim-goto:
The issue with `CMAKE_OSX_DEPLOYMENT_TARGET` should be fixed at its source (the github runner). Simply setting the version to `macos-13` fixes the issue. Also the pytest output contains the [error](https://github.com/tim-goto/cmake-conan/actions/runs/5065328752/jobs/9093791654#step:6:165) (you do have to scroll a bit and the log not being searchable in this view is also quite inconvenient TBH..). You can just cherry pick [this commit](https://github.com/tim-goto/cmake-conan/commit/9b86891ddd85e63307cbf95045996a658afe7a96).

EDIT: The variable is read from `MACOS_DEPLOYMENT_TARGET` [here](https://github.com/Kitware/CMake/blob/master/Modules/Platform/Darwin-Initialize.cmake#L58), you could also just set that in the github runner config I guess

ssrobins:
> The issue with `CMAKE_OSX_DEPLOYMENT_TARGET` should be fixed at its source (the github runner). Simply setting the version to `macos-13` fixes the issue. Also the pytest output contains the [error](https://github.com/tim-goto/cmake-conan/actions/runs/5065328752/jobs/9093791654#step:6:165) (you do have to scroll a bit and the log not being searchable in this view is also quite inconvenient TBH..). You can just cherry pick [this commit](https://github.com/tim-goto/cmake-conan/commit/9b86891ddd85e63307cbf95045996a658afe7a96).
> 
> EDIT: The variable is read from `MACOS_DEPLOYMENT_TARGET` [here](https://github.com/Kitware/CMake/blob/master/Modules/Platform/Darwin-Initialize.cmake#L58), you could also just set that in the github runner config I guess

@tim-goto thanks for the help, I cherry-picked your changes. I plan to look into why the macOS GitHub Actions images are inconsistent with this setting, perhaps I can contribute a fix for it.

@memsharded last thing left on this is to add a test to cover the new os.version scenario, which I will do in a day or so.

memsharded (approved):
Changes look good. 
Let's wait for the tests that cover the new functionality and it can be merged.
Thanks for your contribution!

ssrobins:
> Changes look good. Let's wait for the tests that cover the new functionality and it can be merged. Thanks for your contribution!

Test coverage is added. I use `--fresh` ([more info](https://cmake.org/cmake/help/v3.24/release/3.24.html#command-line)) in the `cmake` command, otherwise `CMAKE_OSX_DEPLOYMENT_TARGET` stays set for the next test.

memsharded (approved):

PR #519, opened 2023-05-24
Feature/reorg

memsharded:
Small project reorg:

- Example in its own folder
- Reduce 2 files to 1 single ``conan_provider.cmake`` file

cc @ssrobins @tim-goto


tim-goto (approved):
lgtm, will rebase #510 after you merge this

franramirez688 (approved):

ssrobins (approved):
Thanks for the change!

Curious if increasing the Conan minimum was necessary for some sort of compatibility change or is the plan to keep it roughly updated to the latest version?

I guess it's nice having an `example` folder to serve as a label for where an example is, but it does require an extra step to cd into the folder to use the cmake command from documentation. I suppose you could update the doc to run the command from the main dir, but `cmake --build` gets a little clunky in my view:

```bash
cd cmake-conan
cmake -B build -S example -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=conan_provider.cmake -DCMAKE_BUILD_TYPE=Release
cmake --build example/build --config Release
```

I don't feel too strongly about it either way.

It's not related to the change, but in my experience, `cmake` always creates the build directory if it doesn't exist, could you just remove `mkdir build`?

memsharded:
By the stability Conan definition, only the latest patch of a given minor release is supported. So the "Minimum required" conan version for something should always be the latest patch. Also yes, we are changing some internal details of the graph-json serialization that we are using, so better force the latest (apparently it didn't break yet, but it might)


>  but it does require an extra step to cd into the folder to use the cmake command from documentation.

I'd say that the extra step is worth to keep things a bit tidier in the repo, and ``cd`` to the example folder is easy to understand by users, to achieve a bit more natural command line, I agree that the root command reads a bit more clunkier and not typical.

PR #520, opened 2023-05-26
Build example on Linux for CI

uilianries:
Use the Github CI to build the example, so we can validate any change there too.

### DIFF
--- a/.github/workflows/cmake_conan.yml
+++ b/.github/workflows/cmake_conan.yml
@@ -33,7 +33,7 @@ jobs:
         with:
           python-version: "3.10"
       - name: Install Conan
-        run: pip install conan pytest && conan --version
+        run: pip install conan && conan --version
       - name: Setup CMake and Ninja
         uses: lukka/get-cmake@latest
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/cmake_conan.yml
```diff
@@ -24,3 +24,27 @@ jobs:
...
+  example:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3.5.2
+      - name: Setup Python
+        uses: actions/setup-python@v4.6.0
+        with:
+          python-version: "3.10"
+      - name: Install Conan
+        run: pip install conan pytest && conan --version
```
memsharded: This example runner won't need pytest?

memsharded (approved):

.github/workflows/cmake_conan.yml
uilianries:
```suggestion
        run: pip install conan && conan --version
```

uilianries: Ops, copy-past failure. Fixed, thanks.

PR #521, opened 2023-05-29
Add Android support

ssrobins:
Add Android support, which involves setting the Conan profile with the following [Android CMake toolchain](https://developer.android.com/ndk/guides/cmake) parameters:

- ANDROID_ABI
- ANDROID_STL
- ANDROID_PLATFORM

...and setting the NDK path from the `ANDROID_NDK_ROOT` environment variable.

I added [CMAKE_FIND_ROOT_PATH](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH.html) so `find_package` would work in a cross-building scenario.

I added test coverage in the smoke test for the four Android architectures. I decided not to add an Android build of the example in CI to not further extended the unit tests. I'm definitely interested in hearing ideas for how we want to deal with this long-term since I'm sure this won't be the last cross-building scenario to be supported.


conan_provider.cmake
```diff
@@ -1,7 +1,26 @@
 set(CONAN_MINIMUM_VERSION 2.0.5)
 
 
-function(detect_os OS OS_VERSION)
+function(detect_arch ARCH)
```
memsharded: This function should probably get the already detected OS as an input argument to decide based on it? 
In the case of ANDROID is more evident, but what if we need to check arch for others like QNX or Mac, which OS detection already contains some logic?

conan_provider.cmake
```diff
@@ -14,15 +33,25 @@ function(detect_os OS OS_VERSION)
...
+            list(GET PLATFORM_LIST 1 _OS_API_LEVEL)
+            message(STATUS "CMake-Conan: android_platform=${ANDROID_PLATFORM}")
+            set(${OS_API_LEVEL} ${_OS_API_LEVEL} PARENT_SCOPE)
+        endif()
     endif()
 endfunction()
 
 
-function(detect_cxx_standard CXX_STANDARD)
+function(detect_cxx CXX_STANDARD LIB_CXX)
```
memsharded: LIBCXX is a different think than the CXX_STANDARD, I think I'd create a ``detect_lib_cxx`` for it.

conan_provider.cmake
```diff
@@ -229,6 +275,10 @@ macro(conan_provide_dependency package_name)
     endif()
 
     get_property(CONAN_GENERATORS_FOLDER GLOBAL PROPERTY CONAN_GENERATORS_FOLDER)
+    list(FIND CMAKE_FIND_ROOT_PATH "${CONAN_GENERATORS_FOLDER}" index)
```
memsharded: This probably deserves a comment here, saying why this is neessary.

memsharded (commented):
Looking very good, just a couple of minor interface suggestions

conan_provider.cmake
ssrobins: That's definitely a good thing to decide on and be consistent with. Do we use `OS` or do we use the global CMake platform variables such as `ANDROID`, `IOS`, `MSVC`, `APPLE`, etc..? I would prefer the CMake platform variables over `OS` for the following reasons:

- Using a variable named `OS` outside of the context of setting the Conan profile's `os` setting might be confusing in cross-compiling scenarios as it might be unclear whether it's referring to the underlying build platform or the target platform
- We're already heavily using global CMake variables to build the Conan profile so why not use the platform variables too?
- `APPLE` would be convenient for macOS and iOS platforms since its architecture settings for both platforms are set in the same CMake variable

I don't have a strong preference though so I'm happy to change it and continue on that pattern for future pull requests.

ssrobins:
Made suggested fixes and posted comments. Still trying to figure out how to fix the Windows test failure, I'll work on it a bit more later today.

conan_provider.cmake
memsharded: I don't have a strong preference either. I was mostly thinking that at the moment ``elseif(${CMAKE_SYSTEM_NAME} STREQUAL "QNX")`` and ``if(${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")``, that is, based on CMAKE_SYSTEM_NAME instead of the CMake variables. 

I am not opposed to change the detection of OS, to make it consistent, but the same ``ifs`` should be applied in the different places. Maybe change that first in a previous PR?

ssrobins: Sure, I'll try that out, probably not until the weekend, it's a busy week.

ssrobins: After thinking more about it, I think we should stick with `CMAKE_SYSTEM_NAME`. The simplicity of setting its value as the value for the `os` Conan setting is going to work for a lot of platforms, including Android and iOS and even exotic platforms with custom toolchains might 'just work'.

So I'll take the route in your original comment:
- All `if(ANDROID)` get changed to `if(${CMAKE_SYSTEM_NAME} STREQUAL "Android")`
- `OS` variable is passed to `detect_arch` and `detect_lib_cxx` and `if(${OS} STREQUAL "Android")` can be used for any Android-specific logic.

Sound good?

memsharded: Yes, exactly!

memsharded (approved):
LGTM, thanks!

memsharded:
Tests are failing in Windows, but not evident why in the CI

ssrobins:
> Tests are failing in Windows, but not evident why in the CI

All I know so far is `pytest` has the failure, but `pytest -k android` does not, which tells me there's some sort of pollution of the test environment by the other tests. I just haven't figured out what it is yet. I'll keep on it.

ssrobins:
> > Tests are failing in Windows, but not evident why in the CI
> 
> All I know so far is `pytest` has the failure, but `pytest -k android` does not, which tells me there's some sort of pollution of the test environment by the other tests. I just haven't figured out what it is yet. I'll keep on it.

Adding a setup step to delete the build directory fixed the problem. Deleting just the conan dir was sufficient to fix it, but I decided to just delete the whole build dir. It might be a good idea to consider a separate effort to have unique build dirs for each test. Separate conan caches might be a good idea too, I think that would also open the door to parallelizing the tests. I kept it simple in this pull request, especially with the current test changes going on in https://github.com/conan-io/cmake-conan/pull/522.

Anyway, this one is ready for merge from my perspective, but let me know if you see anything else that needs addressing.

conan_provider.cmake
```diff
@@ -14,18 +14,51 @@ function(detect_os OS OS_VERSION)
...
+            message(STATUS "CMake-Conan: android_platform=${ANDROID_PLATFORM}")
+            set(${OS_API_LEVEL} ${_OS_API_LEVEL} PARENT_SCOPE)
+        endif()
+    endif()
+endfunction()
+
+
+function(detect_arch OS ARCH)
+    if(${OS} STREQUAL "Android")
+        if(ANDROID_ABI STREQUAL arm64-v8a)
```
CedricSchmeits: Shouldn't you use `CMAKE_ANDROID_ARCH_ABI`? As this is in CMake as of version 3.7.

memsharded:
I think this is ready, assigning it to @jcar87 for a final review.

conan_provider.cmake
ssrobins: From what I've seen, both work to generate the proper build, both check for valid values. It seems slightly more intuitive to use to use `ANDROID_ABI` since it's the [documented input parameter](https://developer.android.com/ndk/guides/cmake). What do you think the advantage would be in using `CMAKE_ANDROID_ARCH_ABI` over `ANDROID_ABI`?

CedricSchmeits: As at that location and at https://cmake.org/cmake/help/latest/variable/CMAKE_ANDROID_ARCH_ABI.html
`CMAKE_ANDROID_ARCH_ABI` is described as the actual CMake variable that is used. But I don't know much of Android builds.

Following when I look at the toolchain for Android:
https://android.googlesource.com/platform/ndk/+/master/build/cmake/android.toolchain.cmake
there are the following lines:
```
# Determine the ABI.
if(NOT CMAKE_ANDROID_ARCH_ABI)
  if(ANDROID_ABI STREQUAL "armeabi-v7a with NEON")
    set(CMAKE_ANDROID_ARCH_ABI armeabi-v7a)
  elseif(ANDROID_ABI)
    set(CMAKE_ANDROID_ARCH_ABI ${ANDROID_ABI})
  elseif(ANDROID_TOOLCHAIN_NAME MATCHES "^arm-linux-androideabi-")
    set(CMAKE_ANDROID_ARCH_ABI armeabi-v7a)
  elseif(ANDROID_TOOLCHAIN_NAME MATCHES "^aarch64-linux-android-")
    set(CMAKE_ANDROID_ARCH_ABI arm64-v8a)
  elseif(ANDROID_TOOLCHAIN_NAME MATCHES "^x86-")
    set(CMAKE_ANDROID_ARCH_ABI x86)
  elseif(ANDROID_TOOLCHAIN_NAME MATCHES "^x86_64-")
    set(CMAKE_ANDROID_ARCH_ABI x86_64)
  else()
    set(CMAKE_ANDROID_ARCH_ABI armeabi-v7a)
  endif()
endif()
```
So also in the toolchain the `CMAKE_ANDROID_ARCH_ABI` is set...

memsharded: Then it would seem that ``CMAKE_ANDROID_ARCH_ABI`` is the more general one, to use both with and without the AndroidNDK toolchain.cmake?

ssrobins: Yeah, that's a compelling reason, I updated it to use `CMAKE_ANDROID_ARCH_ABI`, thanks for the feedback, @CedricSchmeits and @memsharded !

conan_provider.cmake
```diff
@@ -14,18 +14,51 @@ function(detect_os OS OS_VERSION)
         else()
             set(${OS} ${CMAKE_SYSTEM_NAME} PARENT_SCOPE)
         endif()
+        if(${CMAKE_SYSTEM_NAME} STREQUAL "Android")
+            string(REPLACE "-" ";" PLATFORM_LIST ${ANDROID_PLATFORM})
```
jcar87: From the documentation, I can see that the `ANDROID_PLATFORM` can take 3 different formats. Looks like the logic here would work with `android-32` or just `32` - but maybe not for the one where only the letter is specified? I wonder if we need different cases based on regular expressions! 

Not a blocker by any beans if both `android-32` and `32` would both map to the correct value

conan_provider.cmake
```diff
@@ -14,18 +14,51 @@ function(detect_os OS OS_VERSION)
...
+            message(STATUS "CMake-Conan: android_platform=${ANDROID_PLATFORM}")
+            set(${OS_API_LEVEL} ${_OS_API_LEVEL} PARENT_SCOPE)
+        endif()
+    endif()
+endfunction()
+
+
+function(detect_arch OS ARCH)
+    if(${OS} STREQUAL "Android")
+        if(CMAKE_ANDROID_ARCH_ABI STREQUAL arm64-v8a)
```
jcar87: Is `CMAKE_SYSTEM_PROCESSOR` not already correctly set by now? 
The legacy Android toolchain does the following mapping:
* `arm64-v8a` -> `aarch64`
* `armeabi-v7a` -> `armv7-a`
* `x86` -> `i686`

Only saying this because we could cover more ground (i.e. cross-building on Linux, not just Android) if we use the values from `CMAKE_SYSTEM_PROCESSOR` instead.

conan_provider.cmake
```diff
@@ -110,6 +154,13 @@ function(detect_host_profile output_file)
 
     string(APPEND PROFILE "[conf]\n")
     string(APPEND PROFILE "tools.cmake.cmaketoolchain:generator=${CMAKE_GENERATOR}\n")
+    if(${MYOS} STREQUAL "Android")
+        if(EXISTS $ENV{ANDROID_NDK_ROOT})
+            string(APPEND PROFILE "tools.android:ndk_path={{ os.getenv('ANDROID_NDK_ROOT') }}\n")
```
jcar87: Does _using_ the Android NDK _require_ this path to be defined as an environment variable when invoking CMake - or can cmake be invoked simply by passing the absolute path to the android toolchain file?

Because I would expect that at this stage, if the Android toolchain has already been processed, the following variable:
https://cmake.org/cmake/help/latest/variable/CMAKE_ANDROID_NDK.html points to the correct path - which would mean an env var is not strictly required.

conan_provider.cmake
jcar87: mmm I'm not entirely sure about this one, is this 100% needed?

The `CONAN_GENERATORS_FOLDER` would only contain files generated by `CMakeDeps` to resolve `find_package` calls.  Ensuring that `CMAKE_FIND_ROOT_PATH` needs to be looked at first, would only be needed if loading the Android toolchain causes CMake to look for packages there _first_, possibly finding dependencies outside of the intended Conan packages.

However, does that actually happen? I cannot find any `.cmake` finds inside the android NDK that would satisfy any `find_package` call - at least not in NDK version r25b which is the one I'm looking at.

If this is indeed needed because there are scenarios in which the wrong package is found - I would probably suggest taking advantage of the fact that `find_package` is relegated to the dependency provider - thus we can alter the search order when invoking `find_package` rather than rely on global flags.

I think the issue may be perhaps that the Android toolchain (at least the legacy file), sets this:
 `  set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)` - it's a bit odd, since there are no files inside the NDK that can satisfy calls to `find_package` - I would propose a different way of doing this, but I'm happy to handle this in a separate PR.

jcar87 (commented):
Just some comments! Don't think any are blockers to merge this, but I do have questions about `CMAKE_FIND_ROOT_PATH` as I'd like to avoid modifying that variable if possible.

conan_provider.cmake
ssrobins: I pushed a fix that will accept `android-32` or `32`. Also, the toolchain changes the letter identifiers to the numbers so we don't have to deal with them. I added some test coverage and spot-checked and it looks better than it was, though it might be worth a closer look and more testing eventually.

conan_provider.cmake
ssrobins: Yes, they are set as you described so I agree, this would be a more general way to set Conan's `arch` for more than just Android. I can do that here or in a new PR.

ssrobins:
@jcar87 not sure why I can't directly reply to your `CMAKE_FIND_ROOT_PATH` comment, but here's my response:

When I don't have `CMAKE_FIND_ROOT_PATH` set, I get this error in the example:

```
CMake Error at /Users/steve/Code/cmake-conan/conan_provider.cmake:291 (find_package):
  By not providing "Findfmt.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "fmt", but
  CMake did not find one.
```

So it's not that it's finding the wrong package, it's just that it's not finding anything for some reason. This is consistent with my experiences in other codebases as well, but only for cross-compiling. I can look into this in a separate PR, I've long wondered why `CMAKE_MODULE_PATH` is not enough!

conan_provider.cmake
ssrobins: No, it doesn't, I should have dug deeper. I switched to using the `CMAKE_ANDROID_NDK` variable instead. That variable gets set by the Android toolchain and it's one of the [documented ways](https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-for-android-with-the-ndk) to define the NDK path without the toolchain.

ssrobins:
Thanks @jcar87 for the wonderful feedback, I made what I think are the critical fixes and will have time to work on the rest of it in about two days, either in this PR if it's not merged or a new one.

conan_provider.cmake
ssrobins: When I don't have `CMAKE_FIND_ROOT_PATH` set, I get this error in the example:

```
CMake Error at /Users/steve/Code/cmake-conan/conan_provider.cmake:291 (find_package):
  By not providing "Findfmt.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "fmt", but
  CMake did not find one.
```

So it's not that it's finding the wrong package, it's just that it's not finding anything for some reason. This is consistent with my experiences in other codebases as well, but only for cross-compiling. I can look into this in a separate PR, I've long wondered why `CMAKE_PREFIX_PATH` is not enough!

ssrobins: When temporarily adding `set(CMAKE_FIND_DEBUG_MODE TRUE)` to the example `CMakeLists.txt` before `find_package()`, it shows all the paths it searches. When the Android toolchain is used, it's not even trying to use the `CMAKE_PREFIX_PATH` locations. It just searches inside the Android NDK itself. It's as-if `set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)` is set, but I don't see it in the regular toolchain, only the legacy one as you observed. Perhaps 'ONLY' is the default behavior of a toolchain? Anyway, I stopped setting the `CMAKE_FIND_ROOT_PATH` variable and instead set `CMAKE_FIND_ROOT_PATH_BOTH` inside `find_package()`. Let me know what you think.

conan_provider.cmake
ssrobins: I ended up doing this in this PR, works great!

ssrobins:
Ok, I think I addressed everything, even the 'future PR' things. Let me know if there's anything else. If it's ready, please choose the squash merge option since it's a bit of a mess, otherwise.

jcar87 (approved):
Awesome, thanks so much for your contribution!

memsharded (approved):
Great job, many thanks!

PR #522, opened 2023-06-02
develop2: test single and multi configuration CMake generators on all platforms

jcar87:
Precursor PR so that we can add tests on Windows to cover multiple `CMAKE_MSVC_RUNTIME_LIBRARY` scenarios.

Changes in this PR:
* Refactor use of `rm -rf *`  to `shutil.rmtree()`, since the former is _not_ a command on typical Windows shells (cmd or powershell) - I was getting errors running this on a local Windows development environment. 
* Refactor  tests such that single and multi configuration are tested on both platforms - the differences are in what generator is passed to CMake:
    * Single config, Windows: `Ninja`
    * Single config, macOS/Linux: empty (would be Makefiles)
    * Multi config, Windows: empty (would be Visual Studio)
    * Multi config, macOS/Linux: `Ninja Multi-Config`
* Condense tests - there are now two where there were 3 before, covering the same functionality 
* Refactor to avoid repeated code for identical logic
* On Windows, initialize vcvars via `ilammy/msvc-dev-cmd@v1` github actions integration, otherwise CMake does not find the correct compiler on Windows with single configuration

### DIFF
--- a/.github/workflows/cmake_conan.yml
+++ b/.github/workflows/cmake_conan.yml
@@ -11,6 +11,11 @@ jobs:
           - os: ubuntu-latest
     steps:
       - uses: actions/checkout@v3.5.2
+      - uses: ilammy/msvc-dev-cmd@v1
+        with:
+          arch: amd64
+        if: ${{ matrix.os == 'windows-2019' }}
+        name: setup msvc vcvars
       - name: Setup Python
         uses: actions/setup-python@v4.6.0
         with:
--- a/tests/test_smoke.py
+++ b/tests/test_smoke.py
@@ -89,49 +89,35 @@ def chdir_build_multi():
 
 
 class TestBasic:
-    @windows
-    def test_windows(self, capfd, chdir_build):
+    def test_single_config(self, capfd, chdir_build):
         "Conan installs once during configure and applications are created"
-        run("cmake .. -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=conan_provider.cmake")
-        out, _ = capfd.readouterr()
-        assert all(expected in out for expected in expected_conan_install_outputs)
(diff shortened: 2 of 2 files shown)


.github/workflows/cmake_conan.yml
```diff
@@ -11,6 +11,11 @@ jobs:
           - os: ubuntu-latest
     steps:
       - uses: actions/checkout@v3.5.2
+      - uses: ilammy/msvc-dev-cmd@v1
+        with:
+          arch: amd64
+        if: ${{ matrix.os == 'windows-2019' }}
+        name: setup msvc vcvars
```
memsharded, line 18: I am not so sure about this. Does this automatically setup the VS prompt with the compiler activated? If that is the case, the tests should run without it, it shouldn't be necessary to run in an activated prompt.

jcar87: cmake with the ninja generator (or any other generator that are not the visual studio) require to run in an activated prompt for compiler detection (so that it detects cl.exe). 

If the goal is be able to invoke pytest in a non-activated prompt - this would complicate the test file further than necessary? - the environment needs to be activated upon invocation of the cmake process, and in order to reliable locate the VS installation on a generic machine we'd have to use vswhere and the likes (more logic). 

perhaps we could use Conan to generate the activation script (since Conan has the logic to deal with locating the visual studio installation, and either capture the environment in a session-scope fixture (in the same way we are doing for `tmpdirs`), or if we have located the activation script, modify the `run` method such that it loads it first ahead of running the command (I think it should work since we have `shell=True`, but I'm not sure _which_ shell python uses in this instance on Windows and if that works reliably with powershell/cmd/bash).

memsharded (approved):

PR #526, opened 2023-06-16
feat: set tool path

tim-goto:
Scenario: I want to install build requirements with conan. In order to use them in my CMake files after the first find_package, I have to get the conan install path into CMake context.

### DIFF
--- a/tests/resources/tool/CMakeLists.txt
+++ b/tests/resources/tool/CMakeLists.txt
@@ -1,0 +1,9 @@
+cmake_minimum_required(VERSION 3.24)
+project(MyToolApp CXX)
+
+set(CMAKE_CXX_STANDARD 17)
+find_package(hello REQUIRED)
+find_program(
+  NINJA_EXECUTABLE ninja
+  PATHS ${CONAN_TOOL_PATH} REQUIRED
+  NO_DEFAULT_PATH NO_CMAKE_FIND_ROOT_PATH)
--- a/conan_provider.cmake
+++ b/conan_provider.cmake
@@ -210,6 +210,30 @@ function(conan_install)
         # success
         set_property(GLOBAL PROPERTY CONAN_INSTALL_SUCCESS TRUE)
     endif()
+
+    # set tool path
+    set(CONAN_TOOL_PATH)
+    string(JSON NUMBER_OF_NODES LENGTH ${conan_stdout} graph nodes)
+    math(EXPR NUMBER_OF_NODES "${NUMBER_OF_NODES}-1")
+    foreach(INDEX RANGE ${NUMBER_OF_NODES})
+        string(JSON CONTEXT GET ${conan_stdout} graph nodes ${INDEX} context)
+        if(NOT CONTEXT STREQUAL "build")
+            continue()
+        endif()
+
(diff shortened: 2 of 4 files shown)


tests/resources/tool/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+cmake_minimum_required(VERSION 3.24)
+project(MyToolApp CXX)
+
+set(CMAKE_CXX_STANDARD 17)
+find_package(hello REQUIRED)
+find_program(
+  NINJA_EXECUTABLE ninja
+  PATHS ${CONAN_TOOL_PATH} REQUIRED
```
memsharded, line 8: This sounds a bit against the intended "transparent" integration design.
Maybe it makes sense to add things directly to the path?

This might be related to a Conan feature we have thought a couple of times: generating CMake "env" files instead of shell script or .bat files with the environment definitions. Maybe worth giving it a try as a Conan built-in feature?

memsharded (commented):
Hi @tim-goto 

Thanks for your contribution.
I think it would be worth to try to consider this first as a Conan built-in feature, as the usage of environment vars like PATH things in pure CMake is a relevant use case also outside of CMake-Conan.

tests/resources/tool/CMakeLists.txt
tim-goto: True, it adds implicit manipulation of state. I could not think of any other way to do this. If you think it is better suited as a conan builtin, I can also create a PR there. Would this be a separate `cmake-env` generator?

memsharded: That is a very good question, I don't know what the best interface would be, maybe an opt-in in CMakeDeps, like it shouldn't be necessary to modify all recipes for this to be active. It also depends on how the consumption story looks like.
Maybe start from a quick and dirty proof of concept as a new generator (cleaner to develop and review, and we can think later about the best UI)

memsharded: Yes, of course, if you want to give it a try in Conan, go ahead. Just beware that this is pure experimentation, no guarantees that it can be moved forward, it will depend on some factors like space for future clean evolution, potential risks of being trapped in a local maxima, maintainability, etc. But if you are willing to try, that would be very welcome, of course! 🙂

jcar87:
Hi @tim-goto, thank you for your contribution.

@memsharded  - the proposed logic more or less matches what the `CMakeToolchain` does by setting `CMAKE_PROGRAM_PATH`. This overcomes one of the limitations mentioned in https://github.com/conan-io/cmake-conan/tree/develop2#known-limitations-with-conan-20, where `find_program()` doesn't work. It may not be 1:1 equivalent to what `CMakeToolchain` does, but it is a good approximation.

This is more or less in line with what I had in mind. @tim-goto, my suggestion (and also to echo @memsharded's concerns) is that we should be appending to `CMAKE_PROGRAM_PATH`, rather than create a custom `CONAN_TOOL_PATH`. This would enable running `find_program()` without any additional changes. 

Although one must note that this will not work for any tool that needs to be located "during" a `project()` call (like a compiler, linker, build tool, etc) - but should work for anything that happens after the first `find_package()` call.

PR #529, opened 2023-06-26
Add better mappings for windows subsystems and more CPU architectures

jcar87:
* When `CMAKE_SYSTEM_NAME` is either Cygwin or MSYS2, handle it in such a way that this is mapped to `os=Windows` and `os.subsystem` is defined accordingly.
* Handle the `armv7` and `armv8` architectures adding more names they are known as (e.g. `ARM64` -> Windows, `arm64` -> macOS, etc)

Close https://github.com/conan-io/cmake-conan/issues/537


conan_provider.cmake
```diff
@@ -11,6 +11,12 @@ function(detect_os OS OS_API_LEVEL OS_VERSION)
             set(${OS_VERSION} ${CMAKE_OSX_DEPLOYMENT_TARGET} PARENT_SCOPE)
         elseif(${CMAKE_SYSTEM_NAME} STREQUAL "QNX")
             set(${OS} Neutrino PARENT_SCOPE)
+        elseif(${CMAKE_SYSTEM_NAME} STREQUAL "CYGWIN")
+            set(${OS} Windows PARENT_SCOPE)
+            set(${OS_SUBSYSTEM} cygwin PARENT_SCOPE)
+        elseif(${CMAKE_SYSTEM_NAME} STREQUAL "MSYS2")
```
memsharded: This might need to be ``MSYS``, not ``MSYS2``

memsharded (approved):

PR #534, opened 2023-07-13
Add HCOS to list of supported platforms

bwhitchurch:
Hello,
I have been using conan and cmake to manage my builds and dependencies for projects targeting HCOS (honeycomb OS). I would like to add this name to the list of supported platforms.

### DIFF
--- a/conan.cmake
+++ b/conan.cmake
@@ -96,7 +96,7 @@ macro(_conan_check_system_name)
         if(${CMAKE_SYSTEM_NAME} STREQUAL "QNX")
             set(CONAN_SYSTEM_NAME Neutrino)
         endif()
-        set(CONAN_SUPPORTED_PLATFORMS Windows Linux Macos Android iOS FreeBSD WindowsStore WindowsCE watchOS tvOS FreeBSD SunOS AIX Arduino Emscripten Neutrino)
+        set(CONAN_SUPPORTED_PLATFORMS Windows Linux Macos Android iOS FreeBSD WindowsStore WindowsCE watchOS tvOS FreeBSD SunOS AIX Arduino Emscripten Neutrino HCOS)
         list (FIND CONAN_SUPPORTED_PLATFORMS "${CONAN_SYSTEM_NAME}" _index)
         if (${_index} GREATER -1)
             #check if the cmake system is a conan supported one


memsharded (commented):
Adding changes here without adding them first to Conan default ``settings.yml`` is useless, because Conan will reject it if not there.

To add a new OS the process is first sending it to Conan client. It really only makes sense if it is quite mainstream, otherwise, customizing your own ``settings.yml`` would be enough.
Also, latest Conan is 2.0. This kind of support for new platforms should be targeted for Conan 2.0 directly, not for Conan 1.X (both for Conan client, and for the cmake-conan integration).

Thanks very much!

bwhitchurch:
Okay, thank you very much. I had tried adding the entry to our local `settings.yml` as an option under both `os_build` and `os_target`. This works when we call `conan` and `cmake` separately, but it does not seem to be picked up by this script.

memsharded:
``Okay, thank you very much. I had tried adding the entry to our local settings.yml as an option under both os_build and os_target.```

``os_build`` and ``os_target`` are legacy, and shouldn't be used anymore. Cross-build scenarios should use the 2 profiles, with the "build" profile specified with ``-pr:b=mybuildprofile (can be the default one)``

bwhitchurch:
>  customizing your own settings.yml would be enough.

So where should I add a new supported OS if entries under `os_build`, `os_target`, and `os` sections aren't getting picked up? 
I have a profile for this target already as well. I would appreciate if you could take a quick look and tell me what I have missed. Regardless, even with these configurations setup I still get 

>CMake Error at build/hcos_default/conan.cmake:105 (message):
  cmake system HCOS is not supported by conan.  
Use one of
Windows;Linux;Macos;Android;iOS;FreeBSD;WindowsStore;WindowsCE;watchOS;tvOS;FreeBSD;SunOS;AIX;Arduino;Emscripten;Neutrino

because the `SUPPORTED_PLATFORMS` list is not formed from the information in any of these local configuration files.
[hcos_profile](https://github.com/conan-io/cmake-conan/files/12045152/hcos_profile.txt)

memsharded:
> So where should I add a new supported OS if entries under os_build, os_target, and os sections aren't getting picked up?
I have a profile for this target already as well. I would appreciate if you could take a quick look and tell me what I have missed. Regardless, even with these configurations setup I still get

It should be added to the ``os`` section.

> because the SUPPORTED_PLATFORMS list is not formed from the information in any of these local configuration files.
[hcos_profile](https://github.com/conan-io/cmake-conan/files/12045152/hcos_profile.txt)

Yes, I am not saying that this wouldn't need an addition, but without adding it to Conan ``settings.yml`` first, it is pointless.
Also, this kind of new features and support for 1.X series will typically not be considered. New things should target Conan 2.0 branches (also cmake-conan 2 branch, that will not have any constraint in the OS)

Issue #532, opened 2023-07-13
Cmake configuration fail when using Threads


filssavi:
Hi,

The cmake configuration step fails when using std::threads with the following error message:

```
CMake Error at cmake-conan/conan_provider.cmake:278 (find_package):
  Could not find a package configuration file provided by "Threads" with any
  of the following names:

    ThreadsConfig.cmake
    threads-config.cmake

  Add the installation prefix of "Threads" to CMAKE_PREFIX_PATH or set
  "Threads_DIR" to a directory containing one of the above files.  If
  "Threads" provides a separate development package or SDK, be sure it has
  been installed.
Call Stack (most recent call first):
  /snap/cmake/1299/share/cmake-3.26/Modules/CMakeFindDependencyMacro.cmake:76 (find_package)
  /home/filssavi/anaconda3/lib/cmake/GTest/GTestConfig.cmake:29 (find_dependency)
  cmake-conan/conan_provider.cmake:278 (find_package)
  Testing/CMakeLists.txt:31 (find_package)
```

To reproduce the issue you just have to add the following line to the example code CMakeLists.txt

`
find_package(Threads REQUIRED)
`
Suspecting a bug I bisected the issue and found out that the error started popping up in commit 096c99de9e3cffad877ad4eab7394af08d544c8c.

This is however where I run out of steam, as I am not familiar enough with cmake to trace the issue any further

I am on Linux (x86_64 Ubuntu 22.10) with the latest conan installed.

memsharded:
Hi @filssavi 

Thanks for your question.

Could you please share your ``conanfile`` and the full output, not only the final error message? 
It seems that ``gtest`` is not from a Conan package (``/home/filssavi/anaconda3/lib/cmake/GTest/GTestConfig.cmake:29``) is this the intention?

fklebert:
I am facing the same issue. Here is the conanfile:

```
[requires]
openssl/1.1.1t
sqlite3/3.42.0 
spdlog/1.10.0 
zlib/1.2.13 
zstd/1.5.4
lz4/1.9.4

[generators]
CMakeDeps
```

And here is a little more logs (hope the relevant part is in):
```
[cmake] ======== Installing packages ========
[cmake] fmt/8.1.1: Already installed! (1 of 7)
[cmake] lz4/1.9.4: Already installed! (2 of 7)
[cmake] openssl/1.1.1t: Already installed! (3 of 7)
[...]

PR #536, opened 2023-07-17
Ensure that Find<PackageName>.cmake modules work

ssrobins:
This addresses https://github.com/conan-io/cmake-conan/issues/532.

In the process of supporting `find_package()` for cross-compile scenarios, we accidentally excluded `Find<PackageName>.cmake` modules from the search. According to [CMake docs](https://cmake.org/cmake/help/latest/command/find_package.html#search-modes), only 'module mode' will look for these files, which is only available in the basic command signature of `find_package()`. By using `CMAKE_FIND_ROOT_PATH_BOTH` in `find_package()` we switched to the full command signature.

The fix does set `CMAKE_FIND_ROOT_PATH_MODE_PACKAGE` globally like it did before this [pull request comment](https://github.com/conan-io/cmake-conan/pull/521#discussion_r1226940907). I suppose I could save the value of the variable and put it back to avoid a global influence, if you want.

I added test coverage for this scenario. While investigating, I realized the Android tests were failing if the build folder wasn't already there, making it impossible to run `pytest -k android` so I fixed that too!

### DIFF
--- a/conan_provider.cmake
+++ b/conan_provider.cmake
@@ -275,7 +275,8 @@ macro(conan_provide_dependency package_name)
     if(${index} EQUAL -1)
         list(PREPEND CMAKE_PREFIX_PATH "${CONAN_GENERATORS_FOLDER}")
     endif()
-    find_package(${ARGN} BYPASS_PROVIDER CMAKE_FIND_ROOT_PATH_BOTH)
+    set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE BOTH)
+    find_package(${ARGN} BYPASS_PROVIDER)
 endmacro()
 
 
--- a/tests/resources/cmake_module/CMakeLists.txt
+++ b/tests/resources/cmake_module/CMakeLists.txt
@@ -1,0 +1,5 @@
+cmake_minimum_required(VERSION 3.24)
+project(MyApp CXX)
+
+set(CMAKE_CXX_STANDARD 17)
+find_package(Threads REQUIRED)
--- a/tests/test_smoke.py
+++ b/tests/test_smoke.py
@@ -134,6 +134,20 @@ class TestBasic:
         assert all(expected in out for expected in expected_conan_install_outputs)
 
 
+class TestCmakeModule:
+    @pytest.fixture(scope="class", autouse=True)
+    def cmake_module_setup(self):
+        src_dir = Path(__file__).parent.parent
+        shutil.copytree(src_dir / 'tests' / 'resources' / 'cmake_module', ".", dirs_exist_ok=True)
+        yield
+
+    def test_cmake_module(self, capfd, chdir_build):
+        "Ensure that the Find<PackageName>.cmake modules from the CMake install work"
+        run("cmake .. -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=conan_provider.cmake -DCMAKE_BUILD_TYPE=Release")
+        out, _ = capfd.readouterr()
+        assert "Found Threads: TRUE" in out
+
+
 class TestSubdir:
     @pytest.fixture(scope="class", autouse=True)
     def subdir_setup(self):
--- a/tests/test_smoke.py
+++ b/tests/test_smoke.py
@@ -177,7 +177,8 @@ class TestOsVersion:
 class TestAndroid:
     @pytest.fixture(scope="class", autouse=True)
     def android_setup(self):
-        shutil.rmtree("build")
+        if os.path.exists("build"):
+            shutil.rmtree("build")
         yield
 
     def test_android_armv8(self, capfd, chdir_build):


conan_provider.cmake
```diff
@@ -275,7 +275,8 @@ macro(conan_provide_dependency package_name)
     if(${index} EQUAL -1)
         list(PREPEND CMAKE_PREFIX_PATH "${CONAN_GENERATORS_FOLDER}")
     endif()
-    find_package(${ARGN} BYPASS_PROVIDER CMAKE_FIND_ROOT_PATH_BOTH)
+    set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE BOTH)
```
memsharded, line 278: The provider captures every ``find_package()``, so I don't see how restoring the value after the call would make any difference, so I think it is good as-is.

memsharded (approved):
Many thanks for checking it and contributing this so quickly!

conan_provider.cmake
jcar87: I think it's probably best to leave `CMAKE_FIND_ROOT_PATH_MODE_PACKAGE` untouched, regardless of whether we set it in the call to `find_package`, or before as in this PR.

When cross building, the `BOTH` value gives priority and re-roots the entire search order around `CMAKE_FIND_ROOT_PATH` and `CMAKE_SYSROOT` if defined.  So if the Sysroot contains CMake configuration files, it will look for those FIRST, before looking in the `CONAN_GENERATORS_FOLDER` which is what we want. This is likely to happen on sysroots for Linux distros, like Raspberry Pi or NVIDIA Jetson.


The issue with the Android NDK which motivated this workaround, as reported in this previous PR: https://github.com/conan-io/cmake-conan/pull/521#discussion_r1227625701 - has to do with the fact that the Android NDK toolchain sets this to `ONLY`, which causes find_package to never consider anything outside of the sysroot: https://github.com/android/ndk/issues/517

When it comes to the specific variable `CMAKE_FIND_ROOT_PATH_MODE_PACKAGE`, this is probably a defect/bug or oversight in the Android NDK toolchain: there aren't any packages in the sysroot - so while this approach may be valid for libraries and headers, it is not for CMake packages, unless the expectation is that users add files to the Android sysroot:

```
/opt/android_ndk_r25b/toolchains/llvm/prebuilt/darwin-x86_64/sysroot/usr
$  find . -name "*.cmake"
```
^ (the following returns nothing)


If we set this to `BOTH` - when crossbuilding we would be in essence recreating the problems described in this issue: https://github.com/conan-io/conan/issues/12478 - giving more priority to the sysroot than to Conan.

We also cannot set this to `NEVER`, because it's not entirely uncommon to expect some packages to be provided by the SDK/sysroot.

So taking into consideration that we do indeed capture all calls to `find_package`, we probably need a logic that ensures Conan is searched first, and if the package is not found, then it follows normal procedure. CMake's [documentation](https://cmake.org/cmake/help/latest/command/find_package.html) does suggest an option:
```
find_package (<PackageName> PATHS paths... NO_DEFAULT_PATH)
find_package (<PackageName>)
```

which in our case could be translated to:
```
find_package (<PackageName> ${ARGN} PATHS "${CONAN_GENERATORS_FOLDER}" NO_DEFAULT_PATH)
find_package (<PackageName> ${ARGN})
```

The first call, in essence, prevents CMake from looking anywhere BUT `CONAN_GENERATORS_FOLDER` - and only if it is not found there, would the second call come into play. 

We have a secondary issue, which Im not sure it's been reported yet, but our dependency provider will fail for any dependency where the `package_info()` in the recipe is set to  *ONLY* generate a find_module.

ssrobins: This isn't just an Android toolchain issue, iOS/tvOS/watchOS has the same issue. Before posting this fix, I undid the `CMAKE_FIND_ROOT_PATH_BOTH` setting on https://github.com/conan-io/cmake-conan/pull/528 and got a similar error as we did on Android:
https://github.com/conan-io/cmake-conan/actions/runs/5581115133/jobs/10198913086

From here, I think we just have to try solutions against the existing test coverage like your suggestion above, keep adding any missing test coverage, and see what solution satisfies them all. I'm happy to help with reviewing and testing.

I do think this merge did get us closer to working for all scenarios so, in my opinion it should stay on `develop2` while further improvements are made.

jcar87: The situation for iOS/tvOS and watchOS is exactly the same as Android - that variable is set to `ONLY` and that's what it wouldn't work - https://github.com/Kitware/CMake/blob/master/Modules/Platform/Darwin.cmake#L13-L15 

Interestingly enough, when iOS support was added to CMake, the variable `CMAKE_FIND_ROOT_PATH_MODE_PACKAGE` was not touched at all -  https://github.com/Kitware/CMake/commit/11da882a1293b39ddd054342b1e6f2f3bd1bc934#diff-657a0bc2e0a97ee9c1f85f20e7e4d57dbd40b053642de9c7da5f62dedc81f908 
and it was added later to cover for a scenario where the wrong libraries may be picked up: https://github.com/Kitware/CMake/commit/5f5e3062cf2394f4e28e078817e074892e9c508d 

This highlights an interesting limitation of CMake which is hinted at in https://github.com/conan-io/conan/issues/12478: the CMake defaults for iOS/watchOS/tvOS mean that `find_package` will never actually locate a package in config mode (with a `-config.cmake`) *at all* - But this also points at the solution, either just prepend the generators folder to `CMAKE_FIND_ROOT_PATH` - or invoke find_package twice.


My concern with focusing only on known test coverage is that applying fixes without addressing or understanding the root causes may actually cause breakages in those scenarios that are not currently covered by tests, but are perfectly valid otherwise, like cross-building for a Linux distro when the sysroot contains cmake packages. So if we have enough knowledge to prevent those issues irrespective of the testing level, the better!

ssrobins: Thanks for the additional info!

I said add test coverage not just use existing coverage and I certainly agree on proactively solving problems. I'm even willing to work on it though I'd prefer to finish up the two other pending PRs first, if possible.

jcar87: Oh, thank you for discovering and bringing attention to these issues in the first place! It's a great motivator to move this tool forward :D 

The other two pending PRs are on my priority list this week :)

PR #539, opened 2023-07-19
Fix priorities when invoking find_package from the dependency provider

jcar87:
We have seen issues when invoking `find_package( ... BYPASS_PROVIDER)` from the dependency provider, depending on:
* Whether `CMAKE_FIND_ROOT_PATH` is set by the user (or a toolchain) explicitly
* Whether `CMAKE_FIND_ROOT_PATH_MODE_PACKAGE` is set (it is implicitly by the Android toolchain, and by CMake's iOS/tvOS/watchOS handling) - all values of this variable are problematic, including the default behaviour when it isn't set:
    * `ONLY` means `find_package` would only look for config files in the find root path and will never look inside the generators folder, so all `find_package` calls that should be satisfied by a config file would fail when cross-building
    * `BOTH` means that re-rooted paths are searched _first_, which can be a risk when the sysroots contain cmake package files. This is typically the case when cross-building for a Linux distro, e.g. RaspberryPi, or NVIDIA Jetson
    * `NEVER` means that the sysroot is never looked at, which can also be problematic if there are packages that are _not_ provided by Conan, but the user expects to be found by CMake in the sysroot. This could be the case for closed-source libraries that are part of a cross-building SDK, e.g. CUDA for Jetson - but also what we otherwise consider "system libraries" like OpenGL and so on, provided they have a `-config.cmake` 
    
   
 This PR changes the behaviour of the dependency provider such that:
 * For all `find_package` calls that don't explicitly require a `MODULE`, search **first** in the generators folder and use `NO_DEFAULT_PATH` to prevent CMake from looking anywhere else. We assume that if a dependency is satisfied by Conan, that's what the user wants. This also gives priority to the `CONFIG` if available, mimicking the way the `CMakeToolchain` generator would set `CMAKE_FIND_PACKAGE_PREFER_CONFIG`.
 * If this fist call to `find_package` does not satisfy the search, OR if a `MODULE` was explicitly requested - fall back to CMake's built-in implementation. In this case, we assume that if there is _anything_ that is not satisfied by Conan that must be found somewhere else (a CMake built-in module, something inside the sysroot, or a system library) - then we *do not* alter the search behaviour in any way - with ONE exception: if a `MODULE` is explicitly requested, we make sure that Conan-generated find modules take precedence. 
 
 This avoids any modification of `CMAKE_PREFIX_PATH`, `CMAKE_FIND_ROOT_PATH_MODE*` or similar, and ensures that:
 * if a dependency is provided by Conan, use that
 * if not, fall back to letting CMake find based on the state and configuration provided by the user


conan_provider.cmake
```diff
@@ -258,7 +258,7 @@ function(conan_version_check)
 endfunction()
 
 
-macro(conan_provide_dependency package_name)
+macro(conan_provide_dependency method package_name)
```
memsharded, line 261: Is this ``method`` arg unused?

conan_provider.cmake
```diff
@@ -280,12 +280,20 @@ macro(conan_provide_dependency package_name)
     endif()
 
     get_property(CONAN_GENERATORS_FOLDER GLOBAL PROPERTY CONAN_GENERATORS_FOLDER)
-    list(FIND CMAKE_PREFIX_PATH "${CONAN_GENERATORS_FOLDER}" index)
-    if(${index} EQUAL -1)
-        list(PREPEND CMAKE_PREFIX_PATH "${CONAN_GENERATORS_FOLDER}")
+
+    set(_find_args "${ARGN}")
```
memsharded, line 292: A couple of comments describing what is being done, and a link to the CMake docs explaining the 2 ``find_package()`` calls approach would be useful here.

memsharded (approved):
This looks solid, good job!

conan_provider.cmake
jcar87: Yeah this is unused here as we only accept was method - this was a bit wrong before, as the first argument to the macro is the "method" which in this case should be `FIND_PACKAGE` according to the docs: https://cmake.org/cmake/help/latest/command/cmake_language.html#id2, https://cmake.org/cmake/help/latest/command/cmake_language.html#provider-commands. Since our provider only resolves calls to `find_package`, this argument is unused.

Before this PR this variable didn't contain the package name, but rather all the arguments to find_package (including the package name) - with this change, now `package_name` is now the very first argument to the `find_package` call, and `${ARGN}` contains all other arguments.

conan_provider.cmake
jcar87: Added more comments!

tests/resources/libbye/conanfile.py
```diff
@@ -0,0 +1,56 @@
...
+    def package(self):
+        cmake = CMake(self)
+        cmake.install()
+
+    def package_info(self):
+        self.cpp_info.libs = ["bye"]
+
+        # Deviate from default such that `Findbye.cmake`
+        # is also generated by CMakeDeps on the consumer side
+        self.cpp_info.set_property("cmake_find_mode", "both")
```
jcar87, line 56: In order to test `find_package(bye MODULE REQUIRED)`, I need the recipe to know to generate it !

memsharded (approved):

PR #551, opened 2023-08-29
Handle warning when the dependency provider is set but not used

jcar87:
If a consumer passes both `-DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=conan_provider.cmake` and `-DCONAN_COMMAND=/path/to/bin/conan`, but the project makes no calls to `find_package()`, user will see this warning:
```
CMake Warning:
  Manually-specified variables were not used by the project:

    CONAN_COMMAND
```

This PR makes use of the `DEFER` `cmake_language()` feature such that if at the end processing the top-level `CMakeLists.txt` the dependency provider was not invoked at least once, it displays a more meaningful warning, while at the same time it suppresses the unused variable warning:

```
CMake Warning at conan_provider.cmake:362 (message):
  Conan is correctly configured as dependency provider, but Conan has not
  been invoked.  Please add at least one call to `find_package()`.
  conan_provider.cmake:376 (conan_provide_dependency_check)
  CMakeLists.txt:DEFERRED
```


AbrilRBS (approved):

czoido (approved):

memsharded (approved):
Looks good.
Is it worth a test to cover this functionality? If not difficult, I'd add it.

Issue #552, opened 2023-08-30
find_package(Boost REQUIRED COMPONENTS ..) doesn't work with conan_provider.cmake


temap:
Add the following changes to the example:
```diff
diff --git a/example/CMakeLists.txt b/example/CMakeLists.txt
index dd1fedb..5eab55d 100644
--- a/example/CMakeLists.txt
+++ b/example/CMakeLists.txt
@@ -5,6 +5,7 @@ project(FormatOutput LANGUAGES CXX)
 set(CMAKE_CXX_STANDARD 14)
 
 find_package(fmt REQUIRED)
+find_package(Boost REQUIRED COMPONENTS regex)
 
 add_executable(main main.cpp)
-target_link_libraries(main PRIVATE fmt::fmt)                    
\ No newline at end of file
+target_link_libraries(main PRIVATE fmt::fmt)                    
diff --git a/example/conanfile.txt b/example/conanfile.txt
index e83f3c8..f1905c3 100644
--- a/example/conanfile.txt
+++ b/example/conanfile.txt
@@ -1,5 +1,6 @@
 [requires]
+boost/1.82.0
 fmt/9.1.0
```
 
```diff
 [layout]
-cmake_layout
\ No newline at end of file
+cmake_layout
```

Output from `cmake -S . -B build`:

```
-- The CXX compiler identification is GNU 12.2.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at /usr/share/cmake-3.25/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Boost (missing: Boost_INCLUDE_DIR regex)
Call Stack (most recent call first):
  /usr/share/cmake-3.25/Modules/FindPackageHandleStandardArgs.cmake:600 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-3.25/Modules/FindBoost.cmake:2376 (find_package_handle_standard_args)
  CMakeLists.txt:8 (find_package)


-- Configuring incomplete, errors occurred!
See also "/home/panfilov/cmake-conan/example/build/CMakeFiles/CMakeOutput.log".

```

memsharded:
Hi @temap 

Thanks very much for your report, and the details to reproduce, it really helped.
[...]

PR #553, opened 2023-08-31
Avoid CMake built-in modules to be called when a Conan-provided config is already found

jcar87:
We follow this approach - to ensure that the provider is searched first, and then we fall back to modules and non-Conan provided dependencies.
```
find_package (<PackageName> PATHS paths... NO_DEFAULT_PATH)
find_package (<PackageName>)
```
The CMake documentation suggests this, with:

> Once one of the calls succeeds the result variable will be set and stored in the cache so that no call will search again.

This last event does not appear to be entirely true - if the name of the found package coincides with a CMake built-in Find module, given that the second call is likely to prioritize modules (as per CMake defaults), it appears it calls the module as well. This has been observed in particular when `COMPONENTS` are specified for boost.

Close https://github.com/conan-io/cmake-conan/issues/552


tests/test_smoke.py
```diff
@@ -146,10 +152,27 @@ def find_module_setup(self):
     def test_find_module(self, capfd, chdir_build):
         "Ensure that a call to find_package(XXX MODULE REQUIRED) is honoured by the dependency provider"
         run("cmake .. -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=conan_provider.cmake -DCMAKE_BUILD_TYPE=Release", check=False)
-        out, _ = capfd.readouterr()
+        out, err = capfd.readouterr()
```
jcar87:
```suggestion
        out, _ = capfd.readouterr()
```

tests/test_smoke.py
```diff
@@ -150,6 +156,23 @@ def test_find_module(self, capfd, chdir_build):
...
+        shutil.copytree(src_dir / 'tests' / 'resources' / 'find_module_builtin', ".", dirs_exist_ok=True)
+        yield
+
+    @pytest.mark.parametrize("use_find_components", [True, False])
+    def test_find_builtin_module(self, capfd, use_find_components, chdir_build):
+        "Ensure that a Conan-provided -config.cmake file satisfies dependency, even when a CMake builtin "
+        "exists for the same dependency"
+        boost_find_components = "ON" if use_find_components else "OFF"
+        run(f"cmake .. -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=conan_provider.cmake -DCMAKE_BUILD_TYPE=Release -D_TEST_BOOST_FIND_COMPONENTS={boost_find_components}", check=False)
+        out, err = capfd.readouterr()
```
jcar87:
```suggestion
        out, _ = capfd.readouterr()
```

memsharded (approved):

PR #554, opened 2023-09-05
Added [buildenv] Field Generation

jelin-sh:
Modified the `detect_host_profile` function in `conan_provider.cmake` to support the generation of the [buildenv] field. This enhancement allows for a closer integration with the compilers configured in CMake, thereby improving usability for cross-compilation and cross-platform scenarios.

### DIFF
--- a/conan_provider.cmake
+++ b/conan_provider.cmake
@@ -203,6 +203,11 @@ function(detect_host_profile output_file)
         string(APPEND PROFILE "tools.android:ndk_path=${CMAKE_ANDROID_NDK}\n")
     endif()
 
+    string(APPEND PROFILE "[buildenv]\n")
+    string(APPEND PROFILE "CC=${CMAKE_C_COMPILER}\n")
+    string(APPEND PROFILE "CXX=${CMAKE_CXX_COMPILER}\n")
+    string(APPEND PROFILE "LD=${CMAKE_LINKER}\n")
+
     message(STATUS "CMake-Conan: Creating profile ${_FN}")
     file(WRITE ${_FN} ${PROFILE})
     message(STATUS "CMake-Conan: Profile: \n${PROFILE}")


conan_provider.cmake
```diff
@@ -203,6 +203,11 @@ function(detect_host_profile output_file)
         string(APPEND PROFILE "tools.android:ndk_path=${CMAKE_ANDROID_NDK}\n")
     endif()
 
+    string(APPEND PROFILE "[buildenv]\n")
+    string(APPEND PROFILE "CC=${CMAKE_C_COMPILER}\n")
+    string(APPEND PROFILE "CXX=${CMAKE_CXX_COMPILER}\n")
+    string(APPEND PROFILE "LD=${CMAKE_LINKER}\n")
```
memsharded, line 209: Now Conan provides a ``compiler_executables`` configuration that might be a better mechanism to define this (it can work even with build systems that will not listen to CC/CXX env-vars). Setting the env-vars might not be necessarily bad, but we need to discuss the best approach.

memsharded (commented):
Thanks for your contribution @skjsnb 

We need to check what would be the best, most general approach to this, but I think it is worth to pursue this improvement.

jcar87:
Please note that as of https://github.com/conan-io/cmake-conan/pull/566 - the `compiler_executables` configuration variable is set and propagated - this should negate the need for `CC` and `CXX`. Note that this needs CMake to be aware of _both_ compilers (which is not always the case), so requires both to be specified (either by defining `CMAKE_<LANG>_COMPILER`, or letting CMake find it if it is defined as a language in the `project()` call or an `enable_language()`.

The linker is more tricky - since in a lot of scenarios CMake performs linking via the compiler, and `CMAKE_LINKER` is, to the best of my knowledge, not documented. For cases where it is necessary to define these (via `[buildenv]` or similar, I would suggest looking at the just-merged feature of customizing the profile:
https://github.com/conan-io/cmake-conan/tree/develop2#customizing-conan-profiles

Issue #555, opened 2023-09-06
Release build getting debug generators folder set


etodanik:
My conan setup ends up pointing to debug generators when I set build type to Release. I suspect that it might be something the plugin misunderstands? 

Here is my log.

```
C:\Users\danny\AppData\Local\Programs\CLion\bin\cmake\win\x64\bin\cmake.exe -DCMAKE_BUILD_TYPE=Release -DCMAKE_MAKE_PROGRAM=C:/Users/danny/AppData/Local/Programs/CLion/bin/ninja/win/x64/ninja.exe -G Ninja -DBUILD_TESTING:BOOL=OFF -DBUILD_SHARED_LIBS:BOOL=ON "-DCONAN_COMMAND=C:\Program Files\Conan\conan\conan.exe" -DBUILD_INTEGRATION:BOOL=ON -DBUILD_WITH_MT:BOOL=ON -DCEF_RUNTIME_LIBRARY_FLAG:STRING=/MD -DUSE_SANDBOX:BOOL=OFF -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=C:\Users\danny\CLionProjects\myproject-addon\conan_provider.cmake -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=C:\Users\danny\CLionProjects\myproject-addon\conan_provider.cmake -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=C:\Users\danny\CLionProjects\myproject-addon\conan_provider.cmake -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=C:\Users\danny\CLionProjects\myproject-addon\conan_provider.cmake -S C:\Users\danny\CLionProjects\myproject-addon -B C:\Users\danny\CLionProjects\myproject-addon\cmake-build-release-visual-studio
-- Found: clang-format
-- Found: cppcheck
-- CMake-Conan: first find_package() found. Installing dependencies with Conan
-- CMake-Conan: Checking if a default profile exists
C:\Users\danny\.conan2\profiles\default
-- CMake-Conan: cmake_system_name=Windows
-- CMake-Conan: cmake_system_processor=x86_64
-- CMake-Conan: CMake compiler=MSVC
-- CMake-Conan: CMake compiler version=19.37.32822.0
-- CMake-Conan: [settings] compiler=msvc
-- CMake-Conan: [settings] compiler.version=193
[...]

PR #556, opened 2023-09-06
Use GENERATOR_IS_MULTI_CONFIG global property

jcar87:
`CMAKE_CONFIGURATION_TYPES` contains a list of configurations supported by the generator, so it was previously assumed to only be set automatically by `project()` - However, it can also be set externally via an environment variable.

This PR: 
* use the `GENERATOR_IS_MULTI_CONFIG` global property instead, as it is more direct and robust.
* Add a test that causes `CMAKE_CONFIGURATION_TYPES` to be forcibly given a value after the call to `project()`, but before the first Conan invocation, when the generator is single-configuration. Before this PR, `conan install` would be invoked twice - now it is invoked once.

Close https://github.com/conan-io/cmake-conan/issues/555


memsharded (approved):
A unit test that covers this would be nice, but as this might be breaking for users I am fine with merging this quickly.

jcar87:
The behaviour is the same as before in most cases, but this should be more robust when `CMAKE_CONFIGURATION_TYPES` is set/overridden rather than be picked up by `project()`. It looks like CLion is setting `CMAKE_CONFIGURATION_TYPES` even when the generator is single-config Ninja, causing `cmake-conan` to install both debug and release. That should have been harmless but redundant. It may be a little tricky to replicate the exact behaviour on the CI tests.

PR #558, opened 2023-09-06
Various fixes and tidy-ups

jhol:


conan_provider.cmake
```diff
@@ -76,6 +76,8 @@ function(detect_arch ARCH)
         set(_ARCH x86)
     elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "AMD64|amd64|x86_64" OR CMAKE_OSX_ARCHITECTURES MATCHES x86_64)
         set(_ARCH x86_64)
+    else()
+        set(_ARCH "${CMAKE_SYSTEM_PROCESSOR}")
```
memsharded, line 80: This can be some architecture not defined in Conan, which would give an error saying so. I guess it is better than not defining the architecture at all.

If this is the only functional change, better treat it separately in a different PR.

memsharded (commented):
Thanks @jhol for your contribution.

When a PR contains too many stylistic changes it is better to separate those changes into an independent PR, and then do another PR just for the functional changes. These functional changes might require some tests if possible.

conan_provider.cmake
jhol: Yes, that's the idea. I have a completely non-standard architecture, which requires a custom `settings.yml` to be installed. The best solution seems to be to let cmake-conan pass it through directly in that case.

I'll update the commit message to add more detail.

jhol:
> Thanks @jhol for your contribution.
> 
> When a PR contains too many stylistic changes it is better to separate those changes into an independent PR, and then do another PR just for the functional changes. These functional changes might require some tests if possible.

Ok no problem. I'll come back with separate PRs for the various topic areas.

PR #559, opened 2023-09-06
Add detection for C++ standard library

jcar87:
Detect `compiler.libcxx` based on introspection of compiler macros, using `check_cxx_source_compiles` CMake functionality (and some static asserts). This is a precursor to avoid relying on the _default_ profile for the host profile in the future.

Rough logic:
* If the compiler is `gcc`, assume `libstdc++` and detect whether we are using the C++11 ABI or not
* If the compiler is `AppleClang`, directly assume `libc++`
* If the compiler is Clang, and we are _not_ on Windows, detect:
  - if we are using `libstdc++` by macro introspection, and if we are, whether we are using the c++11 abi
  - if we are using `libc++` by macro introspection, and set `libc++` in that case.
  
Notes about tests, coverage for the following, on Linux:
* gcc with `libstdc++11`
* gcc with `libstdc++` (by forcing the old abi via a global flag in the project)
* clang with `libstdc++11` (default when clang is used)
* clang with `libc++` on Linux

not covered, but would be good:
* `clang` (from LLVM) on `macOS` - it _should_ detect libc++ however.

Additionally:
* propagate the detected compiler via the `tools.build:compiler_executables` conf - although at this stage we may not be in a position to pass the `C` compiler because the C language may not be enabled at all in CMake
*


conan_provider.cmake
```diff
@@ -200,6 +264,7 @@ function(detect_host_profile output_file)
 
     string(APPEND PROFILE "[conf]\n")
     string(APPEND PROFILE "tools.cmake.cmaketoolchain:generator=${CMAKE_GENERATOR}\n")
+    string(APPEND PROFILE "tools.build:compiler_executables={\"cpp\":\"${CMAKE_CXX_COMPILER}\"}")
```
memsharded: Probably add ``"c"`` compiler too?

conan_provider.cmake
```diff
@@ -89,11 +89,75 @@ function(detect_cxx_standard CXX_STANDARD)
     endif()
 endfunction()
 
+macro(detect_gnu_libstdcxx)
```
memsharded, line 92: I am a bit afraid that this detection could be a bit slow. Are we really sure that this is at least cached, so it doesn't execute at every build, but just when CMake is really modified/reconfigured?

memsharded (approved):
Good!

conan_provider.cmake
jcar87: According to the cmake documentation:

> The check is only performed once, with the result cached in the variable named by <resultVar>. Every subsequent CMake run will re-use this cached value rather than performing the check again, even if the code changes. In order to force the check to be re-evaluated, the variable named by <resultVar> must be manually removed from the cache.


This should be fairly fast as well, but happy to measure the time if needed.

PR #561, opened 2023-09-07
Various tidy-ups

jhol:


jcar87:
Hi @jhol, thank you for your contribution.

Happy with these fixes:
`if(${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")` -> `if(CMAKE_SYSTEM_NAME STREQUAL "Darwin")`

However a lot of the internal variables we are using are lower-case on purpose. While I understand other projects use different conventions (in particular when variables are of a public scope) - this is our stylistic choice. Most examples in the Variables section of [Professional CMake: A Practical Guide](https://crascit.com/professional-cmake/) also use lowercase rather than all upper case.

jhol:
The issue I have is there is a complete mix of styles of variable names:

  * `foo_bar`
  * `FOO_BAR`
  * `MYFOO_BAR`
  * `_FOO`
  * `_BAR`

I am happy for the style to follow any convention, I just want it to be consistent to make it easier to add later patches. Would you be happy for every variable to be styled `foo_bar`? Both local and global variables, and function arguments?

Issue #540, opened 2023-07-20
[develop2] conanfile.py with CMakeDeps in generate() method causes errors


jcar87:
When we have a `conanfile.py` (as opposed to a `conanfile.txt`), and `CMakeDeps` is instantiated in the `generate()` method (as opposed to as the `generators`) attribute, CMake configuration fails with the following error:

```
ERROR: conanfile.py: Error in generate() method, line 20
	deps.generate()
	ConanException: CMakeDeps is declared in the generators attribute, but was instantiated in the generate() method too. It should only be present in one of them.
CMake Error at conan_provider.cmake:386 (message):
  Conan install failed='1'
Call Stack (most recent call first):
  conan_provider.cmake:478 (conan_install)
  CMakeLists.txt:6 (find_package)
```

rturrado:
I've been using the Conan plugin for CMake. **Great job by the way! 🥳 So much needed and appreciated.** 👏

In my case, I had to remove the two following lines in the `generate` function of the `conanfile.py` for the plugin to work:

```
deps = CMakeDeps(self)
deps.generate()
```

But that made my conan builds from the command line, outside CLion, to start failing. Adding those two lines back got my conan builds from the command line back working.

memsharded:
Thanks for the report @rturrado 

What was exactly the error that made you remove the lines from ``generate()``? Maybe the one saying that the generator is "repeated"?

rturrado:
> Thanks for the report @rturrado 
> 
> What was exactly the error that made you remove the lines from ``generate()``? Maybe the one saying that the generator is "repeated"?

Yes. I can post it exactly when I'm at my computer later, but it was saying something like: you cannot define this in two places, you already defined it at the generators...

memsharded:
Ok, that makes sense thanks!
This seems an overlook of the "package creation" flow, as ``cmake-conan`` is very focused on the "consumer" story, but this should be fixed. I am submitting https://github.com/conan-io/cmake-conan/pull/564, as a possible solution, but we might want to discuss the approach, can't guarantee it will be merged as-is, but we will try to look for a solution in the following days.

PR #564, opened 2023-09-17
CMakeDeps should be defined in conanfile

memsharded:
Close https://github.com/conan-io/cmake-conan/issues/540

The ``CMakeDeps`` in command line will be conflicting with other ``CMakeDeps`` declarations in ``conanfile`` files, which might be necessary for conanfile creating packages (that need to be run without cmake-conan integration later)


AbrilRBS (approved):

memsharded:
Closed, @jcar87 will do a better approach, that solves https://github.com/conan-io/cmake-conan/issues/540 without necessarily breaking existing users without CMakeDeps in their conanfile.txt

PR #566, opened 2023-09-20
Propagate compiler executable via profile configuration

jcar87:
Propagate the compiler executables via the `tools.build:compiler_executable` Conan configuration in the generated profile.

This is particularly necessary when the compiler the CMake build is configured with is _anything other than_ what CMake would auto-detect. This fixes an issue where if not propagated, when building dependencies from source (due to `--build=missing`), Conan might pass flags that are Clang-only, when CMake detects the system default (likely to be gcc).

### DIFF
--- a/conan_provider.cmake
+++ b/conan_provider.cmake
@@ -247,6 +247,25 @@ function(detect_host_profile output_file)
 
     string(APPEND PROFILE "[conf]\n")
     string(APPEND PROFILE "tools.cmake.cmaketoolchain:generator=${CMAKE_GENERATOR}\n")
+
+    # propagate compilers via profile
+    set(_conan_c_compiler "")
+    set(_conan_cpp_compiler "")
+    if(CMAKE_C_COMPILER)
+        set(_conan_c_compiler "\"c\":\"${CMAKE_C_COMPILER}\",")
+    else()
+        message(WARNING "CMake-Conan: The C compiler is not defined. "
+                        "Please define CMAKE_C_COMPILER or enable the C language.")
+    endif()
+    if(CMAKE_CXX_COMPILER)
+        set(_conan_cpp_compiler "\"cpp\":\"${CMAKE_CXX_COMPILER}\"")
+    else()
+        message(WARNING "CMake-Conan: The C++ compiler is not defined. "
+                        "Please define CMAKE_CXX_COMPILER or enable the C++ language.")
+    endif()
+
+    string(APPEND PROFILE "tools.build:compiler_executables={${_conan_c_compiler}${_conan_cpp_compiler}}\n")
+
     if(${MYOS} STREQUAL "Android")
         string(APPEND PROFILE "tools.android:ndk_path=${CMAKE_ANDROID_NDK}\n")
     endif()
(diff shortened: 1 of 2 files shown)


conan_provider.cmake
```diff
@@ -247,6 +247,25 @@ function(detect_host_profile output_file)
...
+                        "Please define CMAKE_C_COMPILER or enable the C language.")
+    endif()
+    if(CMAKE_CXX_COMPILER)
+        set(_conan_cpp_compiler "\"cpp\":\"${CMAKE_CXX_COMPILER}\"")
+    else()
+        message(WARNING "CMake-Conan: The C++ compiler is not defined. "
+                        "Please define CMAKE_CXX_COMPILER or enable the C++ language.")
+    endif()
+
+    string(APPEND PROFILE "tools.build:compiler_executables={${_conan_c_compiler}${_conan_cpp_compiler}}\n")
```
memsharded: If not defined compilers, maybe better not define the dictionary at all?

conan_provider.cmake
```diff
@@ -247,6 +247,25 @@ function(detect_host_profile output_file)
 
     string(APPEND PROFILE "[conf]\n")
     string(APPEND PROFILE "tools.cmake.cmaketoolchain:generator=${CMAKE_GENERATOR}\n")
+
+    # propagate compilers via profile
+    set(_conan_c_compiler "")
+    set(_conan_cpp_compiler "")
+    if(CMAKE_C_COMPILER)
+        set(_conan_c_compiler "\"c\":\"${CMAKE_C_COMPILER}\",")
```
memsharded: This logic will force all user projects to do C CXX languages or get the warnings? Maybe reduce the messages to STATUS?

conan_provider.cmake
```diff
@@ -247,6 +247,25 @@ function(detect_host_profile output_file)
 
     string(APPEND PROFILE "[conf]\n")
     string(APPEND PROFILE "tools.cmake.cmaketoolchain:generator=${CMAKE_GENERATOR}\n")
+
+    # propagate compilers via profile
+    set(_conan_c_compiler "")
```
memsharded: I'll move this to its own helper ``define_compiler_executables_conf`` or the like.

memsharded (commented):
Looking good

conan_provider.cmake
jcar87: At this stage in the process, one of the two _must_ already be defined for Conan to detect a compiler altogether. It errors out a lot earlier if not: https://github.com/conan-io/cmake-conan/blob/develop2/conan_provider.cmake#L107

conan_provider.cmake
jcar87: It doesn't force all users to enable both languages, but it does enforce defining both compilers - there's a distinction there, one can define the `CMAKE_<LANG>_COMPILER` variables without enabling the languages.

While this is unfortunate, I think this is a valid warning, unless we have a different logic and try to guess the other compiler when only one is defined.

If the project only enables C++ and the compiler is clang on Linux - ANY dependency that requires the C language will be built with the system default (gcc). Conan will pass `-stdlib=` and the build of that dependency will fail because gcc does not recognize that option. This does not leave the user with any clue as to what is going on, unfortunately. It think the best course of action is to issue a warning that is more visible than just a status.

memsharded (approved):

PR #567, opened 2023-09-21
Add ability to customize build and host profiles

jcar87:
Add ability to customise what gets passed as profiles (`--profile:host` and `--profile:build`), with the `CONAN_HOST_PROFILE` and `CONAN_BUILD_PROFILE`.

Default values:
* `CONAN_HOST_PROFILE`: `autodetect`
* `CONAN_BUILD_PROFILE`: `default`

If `autodetect` is set - it will use the autodetection from CMake as we have been doing so far for the `develop2` branch.

This will also accept:
* Profile names that are installed in the Conan configuration folder - 
* Path to profile files (that may be anywhere)
* A semi-colon separated list of profiles (to be passed via CLI from left to right, with right having most precedence as per documentation)

Examples:
* `CONAN_HOST_PROFILE`: `clang16`: will pass `--profile:host=clang16` and skip autodetection altogether - expects this profile to exist in the user configuration folder
* `CONAN_HOST_PROFILE`: `default;foobar` will pass `--profile:host=default --profile:host foobar` - expects both to exist with those names
* `CONAN_BUILD_PROFILE`: `default;/path/to/foobar` will pass `--profile:build=default --profile:build=/path/to/foobar`, where the latter is expected to be a text file in that location with the profile contents.

note that it needs to be semicolon (rather than just commas) because it's treated as a list in CMake and easier to operate with.


jcar87:
Small issue in CI where `cwd` returns `C:/Users/RUNNER~1/AppData/Local/Temp/tmpi147kuk1conan_test_dir/build/conan_host_profile` but we need `C:/Users/runneradmin/AppData/Local/Temp/tmpi147kuk1conan_test_dir/build/conan_host_profile`

tests/test_smoke.py
```diff
@@ -288,6 +292,36 @@ def test_propagate_non_default_compiler(self, capfd, chdir_build):
         assert "The C compiler is not defined." not in err
         assert 'tools.build:compiler_executables={"c":"/usr/bin/clang","cpp":"/usr/bin/clang++"}' in out
 
+class TestProfileCustomization:
+    def test_profile_defults(self, capfd, chdir_build):
+        """Test the defaults passed for host and build profiles"""
+        run(f"cmake --fresh .. -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=conan_provider.cmake -DCMAKE_BUILD_TYPE=Release", check=True)
+        builddir = str(Path.cwd()).replace('\\','/')
+        out, _ = capfd.readouterr()
+        assert f"--profile:host={builddir}/conan_host_profile" in out
```
franramirez688: It'd be enough to check that the pattern is there instead of getting the whole path:

```python
        out, _ = capfd.readouterr()
        assert re.search("--profile:host=.*/build/conan_host_profile", out)  # buildidr
```

tests/test_smoke.py
```diff
@@ -288,6 +292,36 @@ def test_propagate_non_default_compiler(self, capfd, chdir_build):
...
+        assert "--profile:build=default" in out
+
+    def test_profile_composed_list(self, capfd, chdir_build):
+        """Test passing a list of profiles to host and build profiles"""
+        run(f'cmake --fresh .. -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=conan_provider.cmake -DCMAKE_BUILD_TYPE=Release -DCONAN_HOST_PROFILE="autodetect;foo" -DCONAN_BUILD_PROFILE="default;bar"', check=True)
+        builddir = str(Path.cwd()).replace('\\','/')
+        out, err = capfd.readouterr()
+        assert f"--profile:host={builddir}/conan_host_profile" in out
```
franramirez688: The same here:

```python
        out, err = capfd.readouterr()
        assert re.search("--profile:host=.*/build/conan_host_profile", out)  # buildidr
```

conan_provider.cmake
```diff
@@ -514,4 +537,8 @@ endmacro()
 
 # Add a deferred call at the end of processing the top-level directory
 # to check if the dependency provider was invoked at all.
-cmake_language(DEFER DIRECTORY "${CMAKE_SOURCE_DIR}" CALL conan_provide_dependency_check)
\ No newline at end of file
+cmake_language(DEFER DIRECTORY "${CMAKE_SOURCE_DIR}" CALL conan_provide_dependency_check)
+
+# Configurable variables for Conan profiles
+set(CONAN_HOST_PROFILE "autodetect" CACHE STRING "Conan host profile")
```
memsharded: The ``autodetect`` name might be a bit confusing for users knowing about ``conan profile detect`` command. Maybe ``cmakedetect``? Don't love it either, ``autodetect`` might be good...

conan_provider.cmake
```diff
@@ -445,15 +462,19 @@ macro(conan_provide_dependency method package_name)
         conan_version_check(MINIMUM ${CONAN_MINIMUM_VERSION} CURRENT ${CONAN_CURRENT_VERSION})
         message(STATUS "CMake-Conan: first find_package() found. Installing dependencies with Conan")
         conan_profile_detect_default()
-        detect_host_profile(${CMAKE_BINARY_DIR}/conan_host_profile)
+        if("autodetect" IN_LIST CONAN_HOST_PROFILE)
```
memsharded: Maybe we want to also condition de ``conan profile detect`` for ``default`` only if ``default`` is included in either of the lists? There are users there explicitly avoiding the default profile.

memsharded (approved):
Looking very good!

conan_provider.cmake
jcar87: maybe just the word `auto`? - so that it's not confused with `conan profile detect`

conan_provider.cmake
jcar87: I considered guarding `conan_profile_detect_default()`, but the "autodetected" profile still does `include(default)`, so it's still needed regardless. There's still a couple of PRs pending to have more confidence the autodetected profile can stand on its own. On the other hand, technically speaking, as per the profile composition logic, I think the current behaviour is equivalent to having `default;autodetected` as the value for `CONAN_HOST_PROFILE`

So maybe we can do in this PR:
* refactor `include(default)` from the autodetected profile 
* set `default;autodetected` as the default value for `CONAN_HOST_PROFILE`
* only call `conan_profile_detect_default()` if `default` is found in either `CONAN_HOST_PROFILE` OR `CONAN_BUILD_PROFILE`
☝️  this should have the same behaviour  as before this PR

In a future PR:
* change the default value of `CONAN_HOST_PROFILE` to `autodetect` (or `auto`) - assuming we can detect all relevant settings from CMake for most cases


what do you think?

tests/test_smoke.py
jcar87: yeah, this is a lot better!

franramirez688 (approved):

conan_provider.cmake
memsharded: Sounds good, I don't think that many users will have a profile named "auto"?

conan_provider.cmake
memsharded: Sounds a perfect plan, I like removing the ``include()`` to start with, and doing it without breaking 👍

conan_provider.cmake
jcar87: Going for `auto-cmake` which I think is less likely than `auto`!

conan_provider.cmake
jcar87: Done! - gone for `auto-cmake` as the auto-generated profile name

PR #571, opened 2023-10-04
Fixes #570: Do not remove CMAKE_MODULE_PATH if already set on find_package()

### LINKED ISSUE #570: Setting CMAKE_MODULE_PATH before first find_package() does not restore value
To test this:
```
list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
message(WARNING "CMAKE_MODULE_PATH (1) is set to: '${CMAKE_MODULE_PATH}'")

find_package(OpenSSL REQUIRED)

message(WARNING "CMAKE_MODULE_PATH (2) is set to: '${CMAKE_MODULE_PATH}'")
```

Patch used in conan_provider.cmake:

```
     # Invoke find_package a second time - if the first call succeeded,
     # this will simply reuse the result. If not, fall back to CMake default search
     # behaviour, also allowing modules to be searched.
+    message(WARNING "<-- Saving CMAKE_MODULE_PATH ('${CMAKE_MODULE_PATH}')")
     set(_cmake_module_path_orig "${CMAKE_MODULE_PATH}")
     list(PREPEND CMAKE_MODULE_PATH "${CONAN_GENERATORS_FOLDER}")
     if(NOT ${package_name}_FOUND)
+        message(WARNING "Finding another package")
         find_package(${package_name} ${ARGN} BYPASS_PROVIDER)
     endif()

+    message(WARNING "--> Loading CMAKE_MODULE_PATH ('${_cmake_module_path_orig}')")
    set(CMAKE_MODULE_PATH "${_cmake_module_path_orig}")
```

Output:
```
CMake Warning at cmake/conan_provider.cmake:504 (message):
  <-- Saving CMAKE_MODULE_PATH ('/home/mrussell/Source/mtx/cmake')
  CMakeLists.txt:44 (find_package)


CMake Warning at cmake/conan_provider.cmake:508 (message):
  Finding another package
  CMakeLists.txt:44 (find_package)


-- CMake-Conan: find_package(PkgConfig) found, 'conan install' already ran
CMake Warning at cmake/conan_provider.cmake:504 (message):
  <-- Saving CMAKE_MODULE_PATH
  ('/home/mrussell/Source/mtx/build-5280-debug/conan/build/Debug/generators;/home/mrussell/Source/mtx/cmake')
  /opt/mtx-venv-5280/lib/python3.8/site-packages/cmake/data/share/cmake-3.26/Modules/FindOpenSSL.cmake:206 (find_package)
  cmake/conan_provider.cmake:509 (find_package)
  CMakeLists.txt:44 (find_package)


CMake Warning at cmake/conan_provider.cmake:508 (message):
  Finding another package
  /opt/mtx-venv-5280/lib/python3.8/site-packages/cmake/data/share/cmake-3.26/Modules/FindOpenSSL.cmake:206 (find_package)
  cmake/conan_provider.cmake:509 (find_package)
  CMakeLists.txt:44 (find_package)


CMake Warning at cmake/conan_provider.cmake:512 (message):
  --> Loading CMAKE_MODULE_PATH
  ('/home/mrussell/Source/mtx/build-5280-debug/conan/build/Debug/generators;/home/mrussell/Source/mtx/cmake')
  /opt/mtx-venv-5280/lib/python3.8/site-packages/cmake/data/share/cmake-3.26/Modules/FindOpenSSL.cmake:206 (find_package)
  cmake/conan_provider.cmake:509 (find_package)
  CMakeLists.txt:44 (find_package)


CMake Warning at cmake/conan_provider.cmake:512 (message):
  --> Loading CMAKE_MODULE_PATH ('')
  CMakeLists.txt:44 (find_package)
```

You can clearly see it is setting the variable to an empty string.
I think this might be because of the extra `find_package()` call in that function.

curlybeast:
This fixes the issue for me.

Close https://github.com/conan-io/cmake-conan/issues/570


memsharded (commented):
Thanks for your contribution @curlybeast 

I don't fully get what is the difference, looks similar behavior to me.
Maybe having a unit test that would fail without the change would clarify it (and a test seems necessary anyway). Do you think you could try to contribute a test too? If not, don't worry, just tell us, and we will try to help with that.

UPDATE: Sorry I hadn't seen https://github.com/conan-io/cmake-conan/issues/570, I have added a reference in this PR message.

I know understand the reasons. Still, that further reveals the importance of adding a test to cover this test for now and to avoid future breakage.

curlybeast:
> I don't fully get what is the difference, looks similar behavior to me.
The CMAKE_MODULE_PATH is a list. All this patch does is:

- Only add the provider path once.
- Only remove the provider path if it's in the list.

Using strings as before means that subsequent runs of that function may change the originally saved string irrevocably, which I believe is what is happening. 

> Do you think you could try to contribute a test too? If not, don't worry, just tell us, and we will try to help with that.
If I find time. But I've also not used the test framework you're using.  Would appreciate some help with that if possible.
I agree a test should be added.

curlybeast:
I've added the beginnings of a test (to the same branch / PR) which replicate the `find_package()` calls I'm using in my project.
I am using `REQUIRED` though and we can't really do this. I also can not reproduce the problem (if I revert the fix) using this test, which is odd. I wonder if one of the modules (my hunch is the OpenSSL module) is doing another `find_package()` that triggers this?

If that is the problem, the question is how to test this effectively without requiring a bunch of pre-installed libraries. Thoughts?

tests/resources/preserve_module_path/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+find_program(CCACHE ccache)
+find_package(OpenSSL)
+find_package(ZLIB)
+find_package(prometheus-cpp)
+
+include(CustomModule)
+hello_world()
+
+add_executable(app main.cpp)
+target_link_libraries(app hello::hello bye::bye)
```
memsharded: The executable build can be removed for the test

tests/test_smoke.py
```diff
@@ -234,6 +234,22 @@ def test_find_module(self, capfd, chdir_build):
...
+        src_dir = Path(__file__).parent.parent
+        shutil.copytree(src_dir / 'tests' / 'resources' / 'preserve_module_path', ".", dirs_exist_ok=True)
+        yield
+
+    def test_preserve_module_path(self, capfd, chdir_build):
+        "Ensure that existing CMAKE_MODULE_PATH values remain in place after find_package(XXX) call"
+        run("cmake .. -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=conan_provider.cmake -DCMAKE_BUILD_TYPE=Release Module -DCMAKE_C_COMPILER=/usr/bin/cc", check=False)
+        out, _ = capfd.readouterr()
+        assert "We should be seeing this warning, if we do the test works" in out
+        run("cmake --build .")
```
memsharded: This line an be removed

tests/resources/preserve_module_path/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+cmake_minimum_required(VERSION 3.24)
+project(MyApp CXX)
+
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/modules)
+
+find_program(CCACHE ccache)
+find_package(OpenSSL)
+find_package(ZLIB)
+find_package(prometheus-cpp)
```
memsharded: This would be the thing to simplify, maybe adding a fake ``openssl-config.cmake`` in another folder and adding it to the CMAKE_PREFIX_PATH?

memsharded:
> I've added the beginnings of a test (to the same branch / PR) which replicate the find_package() calls I'm using in my project.
I am using REQUIRED though and we can't really do this. I also can not reproduce the problem (if I revert the fix) using this test, which is odd. I wonder if one of the modules (my hunch is the OpenSSL module) is doing another find_package() that triggers this?

This is good, exactly the reason why adding the tests are important :)

Indeed do not depend on external packages if possible

It is possible to create simple packages locally, that simulate the openssl package behavior, with an internal find_package or whatever (when using CMakeDeps, this will not happen, not sure about your setup, if you were using openssl in a different way, not via CMakeDeps).

tests/resources/preserve_module_path/CMakeLists.txt
curlybeast: Yea, these are the `find_package()` calls before the problem occurs. 
I'll do some digging in my environment to pinpoint exactly what it is.

jcar87:
Hi @curlybeast - I've sync'ed up the changes with the `develop2` branch, and rebased accordingly. I'll take care of the tests as well, as there are a few more things ! Keep this in mind, if you work on this branch locally you'll have to do `git fetch` followed by `git reset --hard origin/develop`, otherwise you'll may get merge errors.

jcar87 (approved):
LGTM

Thanks @curlybeast - the issue happened in a very specific case where two nested `find_package` calls were satisfied by a CMake FIND module, rather than a config. Only in that scenario was `CMAKE_MODULE_PATH` incorrectly restored. Your solution fixes this.

We discovered a second issue - some recipes can define `cpp_info.builddirs`, which result in paths being prepended to `CMAKE_MODULE_PATH` - those were being lost too. Happy to report your proposed solution also fixes that bug :) 

I've added tests for both situations.

curlybeast:
Thank you for the help on this @jcar87, glad to hear I could help the project and that we found / fixed other related issues too! :)

PR #581, opened 2023-10-30
Various tidy-ups (2)

jhol:
This PR supercedes #561 . It has been rebased on top of `develop2`, and excludes the debated variable name harmonization patch.


jcar87 (approved):
LGTM, thanks for the tidy-up!

conan_provider.cmake
```diff
@@ -130,7 +132,7 @@ endmacro()
 
 
 function(detect_lib_cxx OS LIB_CXX)
-    if(${OS} STREQUAL "Android")
+    if(OS STREQUAL "Android")
```
jcar87: This one here is tricky - with the function being invoked like this:
```
detect_lib_cxx(MYOS MYLIB_CXX)
```
the value of `OS` inside this function is simply the string `MYOS` - so removing the reference (${OS}), causes this to compare the string `MYOS` which `Android`, which obviously fails.

This is the only of the `detect_*` functions that takes an input that is not the name of a variable to write to - so it makes it confusing. 
I've refactored this to simply check in CMake for the current system name being Android, I think that's more in line with the rest of the file.

PR #583, opened 2023-11-07
Minor fix to some asserts in the tests

juansblanco:
Fixed wrong asserts for some tests.
I coudn't test it as tests are only run on darwin configurations, but the change shoud be obvious enough.


memsharded (approved):

czoido (approved):

jcar87 (approved):

jcar87:
Nice catch!! thanks

PR #584, opened 2023-11-07
Fix for CMakeDeps generator being duplicated when using a conanfile.py

juansblanco:
Fix for https://github.com/conan-io/cmake-conan/issues/540
- Issue: Fixed an error when CMakeDeps generator was declared twice using the conanfile.py
- Fix: Now if there's a conanfile.py it won't be passed as argument in the conan-provider.cmake to avoid this issue.
- Added Warning when there's a conanfile with no declared generator
- Added tests to cover possible errors with generators


conan_provider.cmake
```diff
@@ -477,13 +477,18 @@ macro(conan_provide_dependency method package_name)
         construct_profile_argument(_host_profile_flags CONAN_HOST_PROFILE)
         construct_profile_argument(_build_profile_flags CONAN_BUILD_PROFILE)
         get_property(_MULTICONFIG_GENERATOR GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
+        if(EXISTS "${CMAKE_SOURCE_DIR}/conanfile.py")
+            set(generator "")
+        else ()
+            set(generator "-g;CMakeDeps")
```
memsharded, line 500: For the ``conanfile.txt`` I'd try:
- Search for the "CMakeDeps" string in the conanfile.txt
- If the string is not present, then display a warning, saying that it should be there, and this will be changed in the future, and add the generator.

I think that having the generator in the ``conanfile.txt`` can also help to switch to the ``conan install`` + ``cmake`` that might be more typical in CI, for example, so I'd aim for that for the future.

memsharded (commented):
Looking good

conan_provider.cmake
juansblanco: Okey, so the idea is that the CMakeDeps should be declared in the conanfile.txt and we want it to be the standard in the future.
Maybe it would be interesting to have the warning  for the conanfile.py too since there is no warning or specific error if there's no CMakeDeps generator declared. It will just fail when running cmake.

memsharded: There is a valid use case that is users that don't want to use the CMakeDeps generated files but they are using the packaged xxx-config.cmake inside the packages. It would be a valid use case (we might need to add more functionality to the provider and collect the dependencies build-folders to pass them to CMAKE_PREFIX_PATH) that we might want to support in the future. So aiming to making it always a user explicitly defined things seems more flexible and less problematic.

juansblanco:
I added warnings for both conanfile.txt and conanfile.py.
I'm not sure if it's mandatory in conanfile.py or if it can work without having one. It could be changed to error instead of warning if needed.
Also there can be the case where it finds the word in a comment without it being declared.

conan_provider.cmake
```diff
@@ -477,13 +477,27 @@ macro(conan_provide_dependency method package_name)
         construct_profile_argument(_host_profile_flags CONAN_HOST_PROFILE)
         construct_profile_argument(_build_profile_flags CONAN_BUILD_PROFILE)
         get_property(_MULTICONFIG_GENERATOR GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
+        if(EXISTS "${CMAKE_SOURCE_DIR}/conanfile.py")
+            file(READ "${CMAKE_SOURCE_DIR}/conanfile.py" outfile)
+            if(NOT "${outfile}" MATCHES ".*CMakeDeps.*")
```
jcar87, line 490: I think this is potentially risky because `conanfile.py` is still Python, so `CMakeDeps` may be anywhere (in a comment, in an import, etc) and still not be properly instantiated.
On the other hand, if the string is not _ANYWHERE_ (as it is being checked here), then we know this is likely going to be an issue, so maybe the warning is OK

tests/test_smoke.py
```diff
@@ -609,3 +611,34 @@ def test_msvc_x86(self, capfd, basic_cmake_project):
...
+                'generate() method too') in err
+
+    # CMakeDeps generator is not declared in either place
+    def test_no_generator(self, capfd, basic_cmake_project):
+        source_dir, binary_dir = change_conanfile(basic_cmake_project, 'no_generator')
+        run(f'cmake -S {source_dir} -B {binary_dir} -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES={conan_provider} -DCMAKE_BUILD_TYPE=Release', check=False)
+        out, _ = capfd.readouterr()
+        assert 'Could NOT find hello (missing: hello_DIR)' in out
+
+def change_conanfile(basic_cmake_project, gen_resource):
```
jcar87: May be a nit pick of mine, but my intuition would have expected this to be declared before the first invocation.

on the other hand I'm not to keep on passing a fixture as a function argument (rather than as a fixture). 

I would suggest:
* Make this this a static function inside the `TestCMakeDepsGenerators` class
* Take the `source_dir` as the argument, and pass it from the calling side

tests/test_smoke.py
```diff
@@ -609,3 +611,34 @@ def test_msvc_x86(self, capfd, basic_cmake_project):
...
+    source_dir, binary_dir = basic_cmake_project
+    os.chdir(source_dir.as_posix())
```
jcar87: the `basic_cmake_project` fixture will be invoked when entering the test, which already does a `chdir` to the build directory for the duration of the test.
This line is changing the cwd to the source directory (so, overriding the fixture's intention), but also in a way that is persisted across the session (there's no yield or anything restoring it back to the original value). 

`source_dir` should already be a full path in the call to `shutil.copy2` so perhaps unnecessary to do a chdir

jcar87 (commented):
Good job! I've rebased the branch on top of the most recent changes on `develop2` -

Added a few comments, but nothing major

tests/test_smoke.py
```diff
@@ -40,6 +40,8 @@ def run(cmd, check=True):
 @contextmanager
 def chdir(folder):
```
jcar87: I need to remove this completely (will do on a separate PR) as it is no longer used

conan_provider.cmake
juansblanco: Yes, it's what I noted in my comment. I still think it may be worth to keep it as it won't show whent it shouldn't and currently is not showing at all, so this can cover most of the cases. But I could understand not wanting it because it's incosistent.

tests/test_smoke.py
juansblanco: Right, I'll give it another look at the fixtures!

tests/test_smoke.py
```diff
@@ -36,7 +36,7 @@
 def run(cmd, check=True):
     subprocess.run(cmd, shell=True, check=check)
 
-
+
```
juansblanco:
```suggestion

```

jcar87 (approved):

PR #587, opened 2023-11-09
Prevent variable overwriting in nested calls to dependency provider

jcar87:
We can have nested calls to `find_package`, which will be handled by `conan_provide_dependency`, which is a macro.

Any variable that we set in `conan_provide_dependency` will be visible by nested calls - if nested calls use the same variable names further on, then the values are rewritten.
As it turns out, this is only causing issues currently as reported here: https://github.com/conan-io/cmake-conan/issues/570 - note that this PR *does not* fix that issue.

This PR does not change behaviour - just stylistic changes that make things more evident.

In this PR:
* Use lowercase local-scope variable names for values retrieved from global properties (stylistic - but avoids confusion with the global properties and enforces the notion that they are local scope variables)
* `unset()` some variables right after we're done using them - before any possible nested call to `find_package` - not that for these variables, everything was working correctly before this PR, but it was coincidence
* Out of abundance of caution, do not reuse the `_find_args` variable name, but rather use a variable that is suffixed with the package name being requested - and `unset()` it immediately after use. The behaviour before this PR was still correct (since the variable was not used after the call to find_package)
* Only mutate `CMAKE_MODULE_PATH` if we need to fall back to CMake default search behaviour - note that we're still affected by https://github.com/conan-io/cmake-conan/issues/570 and the variable _is_ overwritten if we have two nested calls that end up in the fallback block. This will be soon addressed in https://github.com/conan-io/cmake-conan/pull/571/files#diff-d0a3014e65e75c3075c33559a7b0cccb2baaac5e2d2beb52c7884f5bc15bda5f


memsharded (commented):
Looks necessary clean up, good catch

memsharded (approved):

PR #591, opened 2023-11-14
Add detection for MSVC update

Nekto89:


memsharded (commented):
Hi @Nekto89 

However, it is not possible to merge it in the current state. It goes against the default Conan binary compatibility, that is assuming all updates are binary compatible. ConanCenter uses this assumption, so that means this change will never find binaries in ConanCenter, and this is not possible.

We will be considering a different approach to be able to do this, as opt-in, passing some parameter or the like. Please be patience, we will be having a look and think about it.

In the meantime, note that this is just a file that can go in your repo, you can perfectly use your own fork of the provider for your projects. Thanks!

Nekto89:
> Hi @Nekto89
> 
> Thanks very much for your contribution. However, it is not possible to merge it in the current state. It goes against the default Conan binary compatibility, that is assuming all updates are binary compatible. ConanCenter uses this assumption, so that means this change will never find binaries in ConanCenter, and this is not possible.
> 
> We will be considering a different approach to be able to do this, as opt-in, passing some parameter or the like. Please be patience, we will be having a look and think about it.
> 
> In the meantime, note that this is just a file that can go in your repo, you can perfectly use your own fork of the provider for your projects. Thanks!

As far as I know assumption about binary compatibility is wrong. See
https://learn.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=msvc-170#restrictions
So if CCI provides prebuilt binaries that were built with latest and greatest VS 2022 Update 8, it is not safe to link it to application that is being built with VS2022 Update 7 for example. But it would work if prebuilt binaries were built with Update 7 and application uses Update 8 (and only if it's not static libraries and /GL /LTCG aren't used).

I already have fork of this script that allows multiple calls by generating conanfile.py each time. I just thought that this might be useful to everyone.

memsharded:
> As far as I know assumption about binary compatibility is wrong. See
https://learn.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=msvc-170#restrictions
So if CCI provides prebuilt binaries that were built with latest and greatest VS 2022 Update 8, it is not safe to link it to application that is being built with VS2022 Update 7 for example.

The assumption works very good in practice assuming moving only forward and to the latest updates, which is often the recommended practice anyway. Otherwise, ConanCenter should be building x10 the binaries, which is unfeasible. 

The feature might still be valid for other users, this is why I didn't close the PR, but it just has to be opt-in, not by default. No problem, you can leave it closed at the moment, we will re-open or open new PRs when we discuss how to approach it.

Issue #589, opened 2023-11-14
Building packages using Autotools fails for macOS passing detected CMAKE_<LANG>_COMPILER


czoido:
Hi!

I am encountering an issue when trying to build `libcurl`, which uses Autotools, with cmake-conan on macOS. The problem appears to be related to the absolute paths detected for the C and C++ compilers. This can be replicated with a simple example, as shown here: https://github.com/czoido/cmake-conan-example

```
...
checking for path separator... :
checking for sed... /usr/bin/sed
checking for grep... /usr/bin/grep
checking that grep -E works... yes
checking for ar... /usr/bin/ar
checking for a BSD-compatible install... /usr/local/bin/ginstall -c
checking for gcc... /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc
checking whether the C compiler works... no
configure: error: in `/Users/carlosz/.conan2/p/b/libcue2d65d5f375a7/b/build-release':
configure: error: C compiler cannot create executables
See `config.log' for more details
...
```

When running this example on macOS, `CMAKE_CXX_COMPILER` is detected as `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++`. This seems to cause the build process for libcurl with Autotools to fail. However, if I manually change the compilers to `/usr/bin/g++` and `/usr/bin/cc`, the build completes without any issues.

jcar87:
Interesting!
I'd have to have a look at that config.log, this should be easy to reproduce outside of cmake-conan either way.

My suspicion is that this executable alone: `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++` is not properly configured when invoked directly - only when invoked via `xcrun` does it work properly, presumaly something else gets set up along the way (I think the `SDKROOT` is what xcrun passes) (note that `/usr/bin/g++`, `/usr/bin/clang/`, `/usr/bin/cc` and similar are all run via xcrun).

CMake most likely derives the compiler flags required for it to work, and thus can bypass xcrun completely.

Will look into this !

czoido:
Thanks a lot. I can copy the relevant part of the failing log:

```
configure:4917: checking whether the C compiler works
[...]

PR #592, opened 2023-11-17
Do not pass compiler executable for AppleClang if it is the default

jcar87:
While the compiler resides in `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc`, and that is the path that CMake registers, this cannot be invoked directly _unless_ ones passes flags to locate the sysroot with the library etc. CMake does pass this additional information, but other tools downstream might not (e.g. Autotools, as reported in an issue). This is otherwise done automatically when invoking via `/usr/bin/cc`, which in turn invokes it via `xcrun`, defining `SDKROOT` properly and thus it "just works". 

The fix is to detect if the compiler CMake has detected is the default as found by `xcrun` - in that case,  we don't propagate the compiler, and assume that it is the one that will be detected by default by downstream projects.

Close https://github.com/conan-io/cmake-conan/issues/589


memsharded (commented):
I'd say is looking good, better avoid the definition of ``compiler_executables``.

jcar87:
Additionally and come to think of it - when the CMake generator is "Visual Studio", we don't need to propagate compilers either - these are either derived by CMake itself, or Conan activates vcvars on the fly for command line generators. Less a problem, as it is redundant, but worth taking into account.

czoido (approved):

memsharded (approved):

PR #593, opened 2023-11-17
Inject in the PATH the folder from the CMake exe that invoked the provider if cmake is not found in the PATH

czoido:
Related to: https://github.com/conan-io/conan-center-index/issues/21068

The issue arises with the CLion plugin concerning the invocation of CMake. CLion calls CMake with a command like `/Applications/CLion.app/Contents/bin/cmake/mac/bin/cmake -DCMAKE_BUILD_TYPE=Debug ...`, but it does not include the invoked CMake in the environment in any manner. This leads to a problem: if CMake is not installed system-wide, the plugin fails. If there is a system-wide CMake installation, that version is used instead of the one bundled with CLion. Although using a different CMake version is not inherently problematic, it could lead to inconsistencies.

The proposed solution is to check if CMake is in the path and if it's not, then inject to the PATH the same cmake executable that was used to invoke the provider, then Conan will find that binary and proceed with the installation of packages.

I had to remove the test, apparently there's something that Github Actions does not like about the test and won't pass. I tested it locally and apparently it works file, maybe we can find an alternative way of testing this.


tests/test_smoke.py
```diff
@@ -37,6 +37,7 @@ def run(cmd, check=True):
     subprocess.run(cmd, shell=True, check=check)
 
 
+
```
czoido:
```suggestion
```

memsharded (approved):
LGTM, simpler and also the most expected behavior that users will want.

jcar87:
> Although using a different CMake version is not inherently problematic, it could lead to inconsistencies.

Any potential examples on the potential inconsistencies? 

We could have a scenario where:
* CLion invokes `cmake-conan`, which invokes `conan install --build missing` and builds the dependencies
* A CMake installation _is_ available system-wide, but these dependencies are built with CLion's bundled cmake (as per this PR)
* There is an issue building one of the dependencies, and the user attempts to debug this by calling conan install or conan create themselves, directly from the command line.

In the scenario above, the users commands would run with a different version of CMake than the one being propagated by `cmake-conan`.  - This this happened to me I would certainly find it very confusing. I don't think developers are aware that CLion has its own bundled CMake configured by default? I could be wrong - but I would keep this in mind.

Either way I think I would constrain this a little bit, and only propagate the modified `PATH` environment variable _if_ the directory containing CMake is not already there. The reason for this is that _if_ Cmake is alread in the PATH, we are still reordering the entires of `PATH`, and the directory that contains CMake will now be "first". If that directory contains other things, that may be overriding the location of the tools in the system, if there were tools in other places.

conan_provider.cmake
```diff
@@ -388,12 +388,22 @@ function(conan_install)
     # Invoke "conan install" with the provided arguments
     set(CONAN_ARGS ${CONAN_ARGS} -of=${CONAN_OUTPUT_FOLDER})
     message(STATUS "CMake-Conan: conan install ${CMAKE_SOURCE_DIR} ${CONAN_ARGS} ${ARGN}")
+
+
+    # We inject to the PATH the folder where the CMake that invoked the provider lives
+    # so that in case it's not on the system path the conan install does not fail
+    set(_OLD_PATH $ENV{PATH})
+    set(ENV{PATH} "${CMAKE_DIR}:$ENV{PATH}")
```
jcar87: Re: my other comment

We could be on a linux system where cmake was installed in `/usr/bin/cmake` 
The default PATH on a typical Linux is : `/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`

A regular user may also have things in `$HOME/.local/bin` and similar, or an active `venv` in python.

This line here would place `/usr/bin` at the beginning on the list for conan and below (anything conan invokes) - which could be a problem IF there are things that are installed in `/usr/bin` that are also in something that has higher priority (`/usr/local` or `~/.local/bin` or similar). 

I would avoid altering this priority unless strictly necessary, and only modify `PATH` if the directory that contains the CMake executable is not already in it

memsharded: > I would avoid altering this priority unless strictly necessary, and only modify PATH if the directory that contains the CMake executable is not already in it

Yes, this sounds reasonable concern. Modern CMake has ``cmake_path()`` helpers, maybe it is not difficult to check this.

jcar87: Looks a bit convoluted, but:
```
find_program(_cmake_program NAMES cmake NO_PACKAGE_ROOT_PATH NO_CMAKE_PATH NO_CMAKE_ENVIRONMENT_PATH NO_CMAKE_SYSTEM_PATH NO_CMAKE_FIND_ROOT_PATH)
```

This is essentially getting CMake to ONLY look in the `PATH` environment variable. If it is found, we don't need to do anything. It will be found by child processes as well.  If it isn't found then we know there's no globally installed CMake and that it is safe to prepend to the path env var.

We may want to check whether the found via find_program is the same as CMAKE_PROGRAM and do something special if it isn't the same

memsharded (approved):

jcar87 (approved):

PR #594, opened 2023-11-20
Allow customization of the --build flag

ssrobins:
Right now, it's hardcoded to `--build=missing`, but I'd like to be able to override that to pass `--build=never` in continuous integration to enforce that all packages much have a binary. I generalized it so you could also pass a list of any `--build` parameters to address other use cases folks may have.


jcar87:
Hi @ssrobins - thanks for your contribution. We have just added a  more generic `CONAN_INSTALL_ARGS` that should allow this customization, see here: https://github.com/conan-io/cmake-conan/pull/612

We are actively avoiding to have specific flags mapped to specific arguments - and this is meant as the main point for additional customization.

If it's not enough to satisfy your use case, please let us know!

ssrobins:
Makes sense, thanks! https://github.com/conan-io/cmake-conan/pull/612 works fantastically!

PR #598, opened 2023-11-27
Add support for CMake version less than 3.24

rymut:
Allow to use `conan_provider.cmake` with older versions of cmake.
Uses undocumented function `_find_package` to override `find_package` function with `conan_provide_dependency`.

The same method can be used to override `find_program`, `find_library`, `find_path` or `find_file` - by defining macro/function which name matches buildin command cmake creates `_` + command_name function/macro for buildin see. https://github.com/Kitware/CMake/blob/2d0b7798dbaa5625496907a97973087be8db5f53/Source/cmState.cxx#L521

### DIFF
--- a/conan_provider.cmake
+++ b/conan_provider.cmake
@@ -1,5 +1,6 @@
 set(CONAN_MINIMUM_VERSION 2.0.5)
 
+
 function(detect_os OS OS_API_LEVEL OS_SDK OS_SUBSYSTEM OS_VERSION)
     # it could be cross compilation
     message(STATUS "CMake-Conan: cmake_system_name=${CMAKE_SYSTEM_NAME}")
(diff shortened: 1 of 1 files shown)


memsharded (commented):
Hi @rymut 

However, I am afraid that this is not something that we could support. Nested CMake calls and undocumented behavior can have its problems, and will cause us extra maintenance, support, debugging, etc. 

CMake 3.24 has been there for a while, the documented and the recommended integration are the 3.24 dependency providers and updating CMake is reasonably easy, we even have Conan packages that can be used for that.

The good news is that maintaining your own for with your own ``provider.cmake`` is simple, so you can still use it from your fork. But we are sorry that we won't be able to merge this.

Thanks very much again for your contribution anyway!

rymut:
Thanks for quick response. 
I am closing pull request until functionallity present from 2017 will be finally documented in CMake.

Issue #597, opened 2023-11-23
[develop2] Design for further 'conan install' command customization


ssrobins:
I'd like to propose a potential way forward on allowing all sorts of `conan install` command customizations, while not having to duplicate all the flags explicitly in `conan_provider.cmake`. I'm happy to do the work, but wanted to get some alignment on how to go about it first.

I think overall, really common customizations should be explicit options. That's what's already done [for profiles](https://github.com/conan-io/cmake-conan/pull/567). That's also what I'm proposing for build options in https://github.com/conan-io/cmake-conan/pull/594.

The last explicit option I would add is for `--update` like `CONAN_INSTALL_UPDATE`?. When I'm just iterating on code changes, I don't want `--update` set because it takes too long. However, in continuous integration, and in local builds when I know a recipe has been updated, I want to be able to toggle that on in the `cmake` command with a variable.

Beyond that, I think there should be a generic arg variable like `CONAN_INSTALL_ARGS` that will allow you to add one or more other command flags to `conan install`. Here's what I would typically with such a command:

- `--conf tools.system.package_manager:mode=install`: Make sure that any packages with system package dependencies can actually get installed so that the build will work in a clean environment such as continuous integration. For my personal use, this is important for the SDL and SFML packages.
- `--conf tools.system.package_manager:sudo=True`: Make sure system package manager commands are properly escalated. Handy for continuous integration.
- `--conf tools.cmake.cmaketoolchain:generator=Ninja Multi-Config`: SDL's system checks that run during CMake configuration are significantly slower with `Xcode` generator than Ninja or Makefile generators. I'd love to one day get to the bottom of why that is, but until then, I'd like to just use an alternative generator that's different than the one I chose for CMake build.

If instead there is a preference for having explicit options for `--conf:build`, `--conf:host`, and `--conf:all` options, much like we already do for `--profile:host` and `--profile:build`, I would be totally fine for that. I still figure it would be worthwhile to still have a generic argument for lesser used options. I think in my case, if I needed to get fancier than what I've outlined above, I'd probably just use `conan` commands directly.

What do you think? Happy to implement whatever is decided here!

ssrobins:
@memsharded @jcar87 kind ping on this, just in case it got lost in the shuffle. If you're already aware and this just isn't the priority at the moment, I totally understand.

valgur:
[...]

PR #600, opened 2023-11-29
Allow `conan install` customization by CONAN_ARGS variable - conan command arguments

rymut:
Issue tackle similar issue to pull request conan-io/cmake-conan#594 and conan-io/cmake-conan#597 but with broader strokes.

There are a lot of settings that user would want to set (--build flag in case of pull request conan-io/cmake-conan#594) or in my case pointing to conanfile in a project subfolder as per conan 2.0 documentation [Declaring the layout when the Conanfile is inside a subfolder](https://docs.conan.io/2.0/examples/conanfile/layout/conanfile_in_subfolder.html#examples-conanfile-layout-conanfile-in-subfolder).

The issue with current approach & approach from PR conan-io/cmake-conan#594 & conan-io/cmake-conan#597 is that all important settings would need dedicated CONAN_<name> variable. In my approach direct conan command line arguments are transformad and parsed by `conan_install_format_arguments` & `conan_install_parse_arguments` functions.

For example command `conan install` arguments: `conan/conanfile.txt --build=libA --build libB -b libC`
should be passed to cmake as: `-DCONAN_ARGS=conan/conanfile.txt;--build=libA;--build;libB;-b;libC`

Then function call `conan_install_format_arguments(CONAN_LIST "${CONAN_ARGS}")` will transform it to CONAN_LIST list variable: `conan/conanfile.txt;BUILD;libA;BUILD;libB;BUILD;libC`
Then calling `conan_install_parse_arguments(ARG ${CONAN_LIST})` would return similar results to `cmake_parse_argumnets`: 
`ARG_BUILD=libA;libB;LibC` and `ARG_PATH=conan/conanfile.txt`. 

Parser supports all `conan install` arguments so checking & extracting all flags is quite simple.

I wrote this partly because of this part of code: https://github.com/conan-io/cmake-conan/blob/c53fbf58e4afe7cb93cb32e730d6647fc47a5dce/conan_provider.cmake#L401  
This is obviously uninteded call to `cmake_parse_arguments` with invalid arguments, which in turn allows to user set variable CONAN_ARGS and which will be passed without any checks to `conan install` after appending `--output-folder` argument.

### DIFF
--- a/conan_provider.cmake
+++ b/conan_provider.cmake
@@ -396,14 +396,253 @@ function(conan_profile_detect_default)
     endif()
 endfunction()
 
+function(conan_install_format_arguments ARGUMENTS)
+    set(_conan_values
+        -f
+        --format
+        -v
+        --name
+        --version
+        --user
+        --channel
+        --requires
+        --tool-requires
+        -b
+        --build
+        -r
+        --remote
+        -pr
+        --profile
+        -pr:b
+        --profile:build
+        -pr:h
+        --profile:host
+        -pr:a
+        --profile:all
+        -o
+        --options
+        -o:b
+        --options:build
+        -o:h
+        --options:host
+        -o:a
+        --options:all
+        -s
+        --settings
+        -s:b
+        --settings:build
+        -s:h
+        --settings:host
+        -s:a
+        --settings:all
+        -c
+        --conf
+        -c:b
+        --conf:build
+        -c:h
+        --conf:host
+        -c:a
+        --conf:all
+        --lockfile-out
+        --lockfile-overrides
+        -g
+        --generator
+        -of
+        --output-folder
+        -d
+        --deployer
+        --deployer-folder
+    )
+    unset(_args)
+    foreach (_arg IN LISTS ARGN)
+        foreach(_conan_value IN LISTS _conan_values)
+            string(FIND "${_arg}" "${_conan_value}=" _conan_value_pos)
+            if (_conan_value_pos EQUAL 0)
+                list(APPEND _args "${_conan_value}")
+                string(REPLACE "${_conan_value}=" "" _arg "${_arg}")
+                break()
+            endif()
+        endforeach()
+        list(APPEND _args "${_arg}")
+    endforeach()
+    set(_conan_replaces
+        --help -h
+        -h HELP
+        --format -f
+        -v VERBOSE
+        -f FORMAT
+        --name NAME
+        --version VERSION
+        --user USER
+        --channel CHANNEL
+        --requires REQUIRES
+        --tool-requires TOOL_REQUIRES
+        --build -b
+        -b BUILD
+        --remote -r
+        -r REMOTE
+        --no-remote -nr
+        -nr NO_REMOTE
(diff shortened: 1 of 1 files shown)


jcar87:
Hi @rymut - thanks for your contribution, it is very thorough!
Unfortunately it is not in our plans to expose a wider command invocation customization, as it is a significant maintenance burden. One of the problems of the integration for Conan 1.x was precisely this - as you mention, it results in having to maintain an interface to Conan via CMake variables.  

While your proposal avoids this issue by having a single `CONAN_ARGS` - I can see that there is significant logic to sanitise and validate the input, and we still need to keep a list of all possible `conan install` arguments to ensure correct validation is performed. Newer versions of Conan may add new ones that would have to be reflected here as well.

Bear in mind that `cmake-conan` is intended to cover the case where CMake's dependency provider functionality can be used for Conan to be a drop-in replacement of the "default" logic and satisfy calls to `find_package` in the simplest possible way. For some teams it may be beneficial if it is not necessary or desirable for developers to be knowledgable of COnan internals or the interface to the Conan command. If it gets to the point where the Conan install invocation needs to be heavily tweaked, at that point I would avise invoking Conan manually before CMake.

Otherwise, for more specialised custom cases, it is perfectly feasible to host a copy of `conan_provider.cmake` in your repository and make changes that are adapted to your needs. 

But unfortunately, at this time we are not considering exposing this in the interface to the dependency provider.  We'll happily review specific use cases (such as the one presented in https://github.com/conan-io/cmake-conan/issues/597) so that we get an idea of what is needed by the community and we can provide suitable advise or make changes we are comfortable with.

rymut:
> While your proposal avoids this issue by having a single `CONAN_ARGS` - I can see that there is significant logic to sanitise and validate the input, and we still need to keep a list of all possible `conan install` arguments to ensure correct validation is performed. Newer versions of Conan may add new ones that would have to be reflected here as well.

To be real only important necessery logic would be to detect & sanatize: 
1. conanfile path or its absence (default `${CMAKE_SOURCE_DIR}`)
2. output-folder or its absence (default `${CMAKE_BINARY_DIR}/conan}`)
3. absence of build options otherwise use default --build=missing
4. absence of --format options
5. profile flags to set build profile.

When I write it it actually seems like a lot - but parser can be simplied only to find args based on regex to match `--<langname>` and `-<short>` name - only to match a few selected arguments : path, -of, --output-folder, --build, -b.
Passing invalid arguments to - `conan install` would just fail due setting invalid set of argumets.
Log currently shows what command is being invoked (case of point 4-5) - so it is easy to pinpoint error by end user.

Currently user can easly mess up conan install with passing wrongly values to CONAN_HOST_PROFILE or CONAN_BUILD_PROFILE or just meses up the profile configuration.

Still the bug remains with ability to pass `CONAN_ARGS` directly to conan install command.

jcar87:
We have merged https://github.com/conan-io/cmake-conan/pull/612 - meant as minimal way to allow customizing the conan install arguments.

No validation at the CMake level is performed - it's only intended for advanced use users with knowledge of the CLI.

PR #602, opened 2023-12-12
Add ability to install conan configuration files from directory

temap:
Implement custom Conan configuration install directory. Users can now specify a particular directory for Conan config by setting the `CONAN_CONFIG_DIR` variable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ git clone https://github.com/conan-io/cmake-conan.git -b develop2
 
 ### Example project
 
-This repository contains a `CMakeLists.txt` with an example project that depends on `fmt`. 
+This repository contains a `CMakeLists.txt` with an example project that depends on `fmt`.
 
 ```bash
 cd cmake-conan/example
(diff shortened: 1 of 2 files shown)


memsharded (commented):
Hi @temap 

However, I am afraid that this is not aligned with the direction we are aiming for this integration.

Having the dependency provider to execute configuration steps that affects not the current project, but the full Conan home installation can be confusing and problematic in some cases.

This PR implements one case, installing config from a directory. If we open the door to this, there will be all the natural follow up requests, requesting support for url, git, file ``conan config install`` origins, and then we have to keep adding custom CMake variables to implement the ``conan config install`` CLI syntax. This is beyond the scope of the ``cmake-conan`` integration, that aims to install the current project dependencies, in the most possible transparent way.  We have learned this in the past, having to recall and use CONAN_XXX specific variables for this kind of tasks ends up being worse than just calling the Conan native commands.

Also, this open the doors to an endless request, implementation, support and maintenance loop. Why ``conan config install`` can be done here and not ``conan remote add``? And then users requesting ways to add, remove and update remotes automatically with the cmake-conan provider. And this is a rabbit hole, recreating the whole Conan CLI syntax in CMake variables in a more fragile and complicated layer.

I am not saying this automation cannot be implemented in the user side, just that it is not suitable as built-in in this repo. There can be different approaches to cover different user needs:

- CMAKE_PROJECT_TOP_LEVEL_INCLUDES is actually a list, if you want to automate other tasks (not only Conan), you can perfectly add your own files there, and they can perform tasks as needed
- If the repo needs some setup tasks (conan and others), then running a setup script is a common approach.

Even if this cannot be merged, we really appreciate your contribution and your feedback. Specifically, we are willing to learn the possible use cases, the dev flows, etc, because we might be able to address them in different ways, for example with ``.conanrc`` or future ``workspaces`` feature that is planned for 2.X, so all feedback is very useful, please let us know. Thanks very much! :)

Issue #578, opened 2023-10-27
[develop2] Backslash in CONAN_GENERATORS_FOLDER leads to an escape character fail on Windows


Alex-PLACET:
Hi
On Windows, the CONAN_GENERATORS_FOLDER variable, which is an output of a call of conan install, has backslash. It leads to a character escape later in the code

Example:
```
[cmake] -- CMake-Conan: CONAN_GENERATORS_FOLDER=C:\Users\alexi\Documents\Dev\project
[cmake] -- CMake-Conan: CONANFILE=C:/Users/alexi/Documents/Dev/project/conanfile.py
[cmake] -- CMake-Conan: find_package(Threads) found, 'conan install' already ran
[cmake] -- Could NOT find Threads (missing: Threads_DIR)
[cmake] -- Performing Test CMAKE_HAVE_LIBC_PTHREAD
[cmake] CMake Error at C:/Users/alexi/Documents/Dev/project/build/Debug/CMakeFiles/CMakeScratch/TryCompile-abq9g1/CMakeLists.txt:2 (set):
[cmake]   Syntax error in cmake code at
[cmake] 
[cmake]     C:/Users/alexi/Documents/Dev/project/build/Debug/CMakeFiles/CMakeScratch/TryCompile-abq9g1/CMakeLists.txt:2
[cmake] 
[cmake]   when parsing string
[cmake] 
[cmake]     C:\Users\alexi\Documents\Dev\project\build\Debug\conan\build\Debug\generators;C:/Qt/6.6.0/msvc2019_64/lib/cmake/Qt6;C:/Qt/6.6.0/msvc2019_64/lib/cmake/Qt6/3rdparty/extra-cmake-modules/find-modules;C:/Qt/6.6.0/msvc2019_64/lib/cmake/Qt6/3rdparty/kwin
[cmake] 
[cmake]   Invalid character escape '\U'.
```

A fix is to add
`cmake_path(CONVERT ${CONAN_GENERATORS_FOLDER} TO_CMAKE_PATH_LIST CONAN_GENERATORS_FOLDER)`

A pullrequest is ongoing

memsharded:
Hi @Alex-PLACET 

Thanks for the report and your contribution.
It is not fully clear in what situations this is failing. Because even if it has the backslash, this is working, at least for the common and tested use cases.
If possible, it would be amazing to have a test that would fail without this fix, to avoid this for breaking in the future too.

nCore:
[...]

PR #603, opened 2023-12-13
Add test for issue #578

juansblanco:
Test for issue: https://github.com/conan-io/cmake-conan/issues/578

This test checks that the cmake function `try_compile` doesn't give an error when passing a path with `\`.
This only occurred on Windows in some cases, like having a `cmake_layout` set.


tests/resources/try_compile/conanfile.txt
```diff
@@ -0,0 +1,5 @@
+[requires]
+fmt/10.1.1
```
memsharded, line 2: It is recommended to start using explicitly ``CMakeDeps`` generator in the conanfile.txt too.

memsharded (approved):
Looks good

tests/resources/try_compile/conanfile.txt
juansblanco: Done!

jcar87 (approved):

PR #610, opened 2024-01-15
add OUTPUT_FOLDER to install command in old_conan_cmake_install

swissbit-csteuer:
Although the OUTPUT_FOLDER argument was already parsed in old_conan_cmake_install, it was not added to the conan_args. Therefore, the OUTPUT_FOLDER argument had no effect when conan_cmake_run without multi-configuration generators.


jcar87:
Hi @swissbit-csteuer - thanks for your contribution.

This is a PR against the legacy integration for Conan 1.x which we are no longer maintaining. We are actively maintaining the current version for Conan 2.0. If you still require changes for the legacy integration, I would advise keeping a fork with your custom changes.

PR #611, opened 2024-01-16
Add a fallback in case CMAKE_OSX_ARCHITECTURES isn't set

Cogitri:
#528 introduced new behaviour to support iOS, tvOS and watchOS. This also added a reliance on CMAKE_OSX_ARCHITECTURES, which may be unset in some configurations when building for MacOS. In this case, we should still fallback to CMAKE_SYSTEM_PROCESSOR and assume we're trying to build for the current target only.


jcar87 (approved):
Hi @Cogitri  - thanks so much! Indeed we missed this one, thanks for fixing it.

Cogitri:
Thanks for the quick merge! :)

PR #612, opened 2024-01-22
Add CONAN_INSTALL_ARGS variable

jcar87:
- Add `CONAN_INSTALL_ARGS` cmake variable to control arguments passed to conan install - with some caveats, see documentation.


README.md
```diff
@@ -60,10 +60,17 @@ Please note that for the above to work, a `default` profile must already exist.
...
+* `-DCONAN_BUILD_PROFILE="/path/to/profile"`: alternatively, provide a path to a profile file that may be anywhere in the filesystem.
+* `-DCONAN_HOST_PROFILE="default;custom"`: semi-colon separated list of profiles. A compound profile will be used (see [docs](https://docs.conan.io/2.0/reference/commands/install.html#profiles-settings-options-conf)) - compunded from left to right, where right has the highest priority.
+
+### Customizing the invocation of Conan install
+The CMake-Conan dependency provider will autodetect and pass the profile information as described above. If the `conan install` command invocation needs to be customized further, the `CONAN_INSTALL_ARGS` variable can be used. 
+* By default, `CONAN_INSTALL_ARGS` is initialised to pass `--build=missing`. If you customize this variable, please be aware that Conan will revert to its default behaviour unless you specify the `--build` flag.
+* Two arguments are reserved to the dependency provider implementation and must not be set: the path to a `conanfile.txt|.py`, and the output format (`--format`).
```
memsharded, line 71: Maybe recall that the ``--profile`` argument is also reserved to the provider?

memsharded (approved):
Looks perfect

README.md
jcar87: True, although that wasn't my initial intention since the `--profile` flags are/ can be additive and still work/. If they are passed via `CONAN_INSTALL_ARGS` they'll be appended - presumably without causing errors in the invocation itself, although this is very "you better know what you're doing" territory. Presumably you can "blank" the two `CONAN_HOST_PROFILE_*` variables, and handle it all with this flag instead.

On the other hand, the path to the conanfile and the `--format` flags would always be incompatible and cause an error if one tries to define them here.

PR #617, opened 2024-02-01
Search conan executable in the default installer location

nolange:
The official Windows installer [1] by default installs conan to `%ProgramFiles%\Conan\conan\conan.exe`.
Add this subdirectory to the find_package call on Windows.

[1] - https://conan.io/downloads


memsharded (commented):
Thanks for your contribution @nolange 

It is not fully clear why this would be necessary. The Windows installer puts Conan in the path, so it should be found by ``find_program`` without further issues.

nolange:
no, putting stuff in path is a really bad idea, windows will search DLLs there.

if you dont have conan in path,  the script wont work obviously

PR #622, opened 2024-02-27
add the TIClang compiler, which is supported as of cmake 3.29.0.

sotten:
CMake version 3.29 adds the TIClang compiler natively. Since it behaves similarly to the Clang compiler, supporting in conan is easy.
https://cmake.org/cmake/help/latest/release/3.29.html#compilers


conan_provider.cmake
```diff
@@ -239,6 +239,10 @@ function(detect_compiler COMPILER COMPILER_VERSION COMPILER_RUNTIME COMPILER_RUN
         set(_COMPILER "clang")
         string(REPLACE "." ";" VERSION_LIST ${CMAKE_CXX_COMPILER_VERSION})
         list(GET VERSION_LIST 0 _COMPILER_VERSION)
+    elseif(_COMPILER MATCHES TIClang)
+        set(_COMPILER "ticlang")
```
memsharded, line 243: This would be an issue. As the ``ticlang`` is not a built-in ``settings.yml``, it will fail for Conan users. Maybe it should be added first there? Or why not just assimilating it to ``clang``? Is it binary incompatible to ``clang``? Different versioning? A bit of scope and extra info about this compiler would be nice

memsharded (commented):
Thanks for your contribution @sotten

conan_provider.cmake
sotten: the settings.yml is generated by this script, right? https://github.com/conan-io/conan/blob/develop2/conans/client/conf/__init__.py

The TIClang, Texas Instruments (TI) calls it tiarmclang (ARM-CGT-CLANG), is used for embedded systems and is usually executed as cross-compiled. The compiler is based on clang currently clang 16, but uses its own version number (currently 3.2.2, it uses Semver). TI has enhanced the clang with its own features for improved debugging, smaller code size and so on.

I use cmake and the TI compiler to create Conan package with jfrog artifactory. Until now, a lot of manual work had to be done in cmake so that cmake could handle the compiler. Now that it is natively supported, I hope that conan will soon recognize the compiler and I can simplify my process.

This is the product page: https://www.ti.com/tool/ARM-CGT
The documentation for the compiler can be found here: https://software-dl.ti.com/codegen/docs/tiarmclang/compiler_tools_user_guide/
The pull request in cmake can be found here: https://gitlab.kitware.com/cmake/cmake/-/merge_requests/9228

PR #626, opened 2024-03-11
Compute CMAKE_PROGRAM_PATH from returned json

jcar87:
This might work to set this _after_ the first call to `find_package` (as the known limitation). 

The implementation in `CMakeToolchain` is:
```
        build_req = self._conanfile.dependencies.build.values()
        build_bin_paths = []
        for req in build_req:
            cppinfo = req.cpp_info.aggregated_components()
            build_paths.extend(cppinfo.builddirs)
            build_bin_paths.extend(cppinfo.bindirs)
```

This is just an approximation by relying on the json alone.

### DIFF
--- a/conan_provider.cmake
+++ b/conan_provider.cmake
@@ -470,7 +470,7 @@ macro(conan_set_program_path)
         string(JSON _dep_binary GET ${conan_stdout} graph nodes ${_node_number} binary)
         string(JSON _dep_context GET ${conan_stdout} graph nodes ${_node_number} context)
         
-        if(NOT _dep_binary STREQUAL "Cache" OR NOT _dep_context STREQUAL "build")
+        if(_dep_binary STREQUAL "Skip" OR NOT _dep_context STREQUAL "build")
             continue()
         endif()
 
@@ -488,7 +488,6 @@ macro(conan_set_program_path)
         endforeach()
     endforeach()
 
-    message("conan_program_path: ${_CONAN_PROGRAM_PATH}")
     list(REMOVE_DUPLICATES _CONAN_PROGRAM_PATH)
 endmacro()
 
--- a/conan_provider.cmake
+++ b/conan_provider.cmake
@@ -451,8 +451,46 @@ function(conan_install)
     # success
     set_property(GLOBAL PROPERTY CONAN_INSTALL_SUCCESS TRUE)
 
+
+    conan_set_program_path()
+
+    list(PREPEND CMAKE_PROGRAM_PATH ${_CONAN_PROGRAM_PATH})
+    set(CMAKE_PROGRAM_PATH ${CMAKE_PROGRAM_PATH} PARENT_SCOPE)
+
 endfunction()
 
+macro(conan_set_program_path)
+    set(_CONAN_PROGRAM_PATH "")
+
+    message("${conan_stdout}")
+    string(JSON _conan_graph_nodes_length LENGTH ${conan_stdout} graph nodes)
+    math(EXPR _conan_graph_nodes_length "${_conan_graph_nodes_length} - 1")
+    foreach(_node_number RANGE ${_conan_graph_nodes_length})
+
+        string(JSON _dep_binary GET ${conan_stdout} graph nodes ${_node_number} binary)
+        string(JSON _dep_context GET ${conan_stdout} graph nodes ${_node_number} context)
+        
+        if(NOT _dep_binary STREQUAL "Cache" OR NOT _dep_context STREQUAL "build")
+            continue()
+        endif()
+
+        string(JSON _current_node_length LENGTH ${conan_stdout} graph nodes ${_node_number} cpp_info)
+        math(EXPR _current_node_components "${_current_node_length} - 1")
+        foreach(_component RANGE ${_current_node_components})
+            string(JSON _current_component_name MEMBER ${conan_stdout} graph nodes ${_node_number} cpp_info ${_component})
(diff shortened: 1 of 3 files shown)


conan_provider.cmake
```diff
@@ -451,8 +451,46 @@ function(conan_install)
...
 
+
+    conan_set_program_path()
+
+    list(PREPEND CMAKE_PROGRAM_PATH ${_CONAN_PROGRAM_PATH})
+    set(CMAKE_PROGRAM_PATH ${CMAKE_PROGRAM_PATH} PARENT_SCOPE)
+
 endfunction()
 
+macro(conan_set_program_path)
```
memsharded, line 462: Maybe it would be a bit cleaner to do a function and "return" the variable to be used later

conan_provider.cmake
```diff
@@ -451,8 +451,46 @@ function(conan_install)
...
+
+    message("${conan_stdout}")
+    string(JSON _conan_graph_nodes_length LENGTH ${conan_stdout} graph nodes)
+    math(EXPR _conan_graph_nodes_length "${_conan_graph_nodes_length} - 1")
+    foreach(_node_number RANGE ${_conan_graph_nodes_length})
+
+        string(JSON _dep_binary GET ${conan_stdout} graph nodes ${_node_number} binary)
+        string(JSON _dep_context GET ${conan_stdout} graph nodes ${_node_number} context)
+        
+        if(NOT _dep_binary STREQUAL "Cache" OR NOT _dep_context STREQUAL "build")
```
memsharded: "Downloaded" or "Updated" will also be available to use, why ignoring them?

conan_provider.cmake
```diff
@@ -451,8 +451,46 @@ function(conan_install)
...
+        string(JSON _current_node_length LENGTH ${conan_stdout} graph nodes ${_node_number} cpp_info)
+        math(EXPR _current_node_components "${_current_node_length} - 1")
+        foreach(_component RANGE ${_current_node_components})
+            string(JSON _current_component_name MEMBER ${conan_stdout} graph nodes ${_node_number} cpp_info ${_component})
+            string(JSON _current_component_bindirs GET ${conan_stdout} graph nodes ${_node_number} cpp_info ${_current_component_name} bindirs)
+            string(JSON _bindirs_length LENGTH ${_current_component_bindirs})
+            math(EXPR _bindirs_length "${_bindirs_length} - 1")
+            foreach(_bindir_index RANGE ${_bindirs_length})
+                string(JSON _bindir  GET ${_current_component_bindirs} ${_bindir_index})
+                list(APPEND _CONAN_PROGRAM_PATH ${_bindir})
```
memsharded, line 486: Master of CMake! I am happy that I didn't have to write this code😄

memsharded (approved):
I think it is fine, can be moved forward.
Maybe we want to consider at some point having something built-in, a new generator or something that can just generate the information in a way that it can be consumed more easily by the provider?

conan_provider.cmake
jcar87: Oh I see! Yeah I need to add those - the ones that have "skip" still have the fields but they are null, so I need a way to filter out anything that isn't valid. Perhaps this isn't the best field to check if the binary package _is_ in the cache (I'm assuming that Cache, Download _and_ Updated for the `binary` field _ALL_ mean they are in the cache with valid paths?)

memsharded: Isn't it enough to filter out the ones with ``binary=="Skip"``? that should work I think.

jcar87: mmm what would happen if we have different versions of the same dependency (via tool requires), and `tools.graph:skip_binaries=False` ? would there be a chance that we add `bin` dirs for different versions of the samae dependency in this case?

memsharded: You can check the node ``context`` attribute, it will be "host" and "build", you can differentiate it.

jcar87: Can go with the condition "binary is not skip" and "context is build" - hopefully that covers all cases correctly!

memsharded: I think so

conan_provider.cmake
```diff
@@ -470,7 +470,7 @@ macro(conan_set_program_path)
         string(JSON _dep_binary GET ${conan_stdout} graph nodes ${_node_number} binary)
         string(JSON _dep_context GET ${conan_stdout} graph nodes ${_node_number} context)
         
-        if(NOT _dep_binary STREQUAL "Cache" OR NOT _dep_context STREQUAL "build")
+        if(_dep_binary STREQUAL "Skip" OR NOT _dep_context STREQUAL "build")
```
rturrado, line 473: Wouldn't you play safer doing a comparison against a lowercase value?:

```
if(_dep_binary_lowercase STREQUAL "skip" OR NOT _dep_context_lowercase STREQUAL "build")
```

rturrado: Or even defining those `"Skip"`, `"build"`... somewhere as constants. Although that may be overkill. 😄 This type of code always looks to me like a bit like "magic numbers".

jcar87: this part of Conan's json formatter, so it's essentially a public API - if this ever changed in Conan, our tests (in Conan) would fail straight away so we would know that would be a breaking change (minimising the chance of this ever happening - so not sure we can something by doing lowercase).

they could be constants but then we would have more indirection given how CMake expands variables but I'll consider that

rturrado: Thanks for the explanation! And roger that CMake code is by itself complicated enough to introduce further indirections.

PR #629, opened 2024-03-18
Specify build type explicitly

planetmarshall:
Pass the build type explicitly with the install command. The build type of the dependencies always has to match the build type of the project in order for the `ConanDeps` generator to work.


memsharded (commented):
Hi @planetmarshall 

Thanks for your contribution.

This shouldn't be necessary. The value of ``CMAKE_BUILD_TYPE`` is already being used in https://github.com/conan-io/cmake-conan/blob/2022a2a96ba8e600eda245bb0b273d652ec02876/conan_provider.cmake#L264,  ``detect_build_type`` and it will be defined in the profile as ``build_type=<value>``, so passing it again in the command call as ``-s`` would be redundant.

planetmarshall:
> 
> This shouldn't be necessary. The value of `CMAKE_BUILD_TYPE` is already being used in
> 
> https://github.com/conan-io/cmake-conan/blob/2022a2a96ba8e600eda245bb0b273d652ec02876/conan_provider.cmake#L264
> , `detect_build_type` and it will be defined in the profile as `build_type=<value>`, so passing it again in the command call as `-s` would be redundant.

This is an issue when the `CONAN_HOST_PROFILE` variable is used rather than automatically detecting a profile. Example project attached: [cmake-conan-test.tar.gz](https://github.com/conan-io/cmake-conan/files/14967663/cmake-conan-test.tar.gz)

## Steps to reproduce

1. Unpack the test project
2. Run
```
cmake --preset conan-release
cmake --build build/conan-release
```
Build succeeds
```
cmake --preset conan-debug
cmake --build build/conan-debug
```
Build fails due to not finding the `spdlog` include (as a separate issue, this error can be difficult to diagnose but it is because the `CMAKE_BUILD_TYPE` does not match the config files generated for the profile).

3. Workaround

Replace the `conan_provider.cmake` file with the one provided by the PR.

memsharded:
Thanks for the feedback.

> conan-default

I'd suggest not using these names for your own Preset files. This is the preset name of Conan-generated files, but if you use it in your own, there might be some collisions.

I am not fully sure I understand the issue, but I'd say that basically you are trying to use the same preset with the same profile for two different configuration. This would be incorrect, in general, for single-config generators, it is usual to have 2 different configure presets, because the configurations need different flags, different inputs, as in this case, different profiles.

planetmarshall:
> Thanks for the feedback.
> 
> > conan-default
> 
> I'd suggest not using these names for your own Preset files. This is the preset name of Conan-generated files, but if you use it in your own, there might be some collisions.

Sure, that's just illustrative for this example. In this example conan doesn't generate its own presets anyway.

> 
> I am not fully sure I understand the issue, but I'd say that basically you are trying to use the same preset with the same profile for two different configuration. This would be incorrect, in general, for single-config generators, it is usual to have 2 different configure presets, because the configurations need different flags, different inputs, as in this case, different profiles.

Maybe I didn't explain it very well.

* If the configurations were significantly different I would agree that it would be better to have completely different profiles for each configuration, but the scenario here is that the configurations are the same in every respect apart from the `build_type`, so your suggestion would lead to a proliferation of profile files that differ only by `build_type`.

* `build_type` is special, as it translates directly to `CMAKE_BUILD_TYPE` which is used in the generated config files - for example
   ```cmake
   set_property(TARGET fmt::fmt APPEND PROPERTY INTERFACE_INCLUDE_DIRECTORIES
                     $<$<CONFIG:Release>:${fmt_fmt_fmt_INCLUDE_DIRS_RELEASE}>)
   ```
   So if the `CMAKE_BUILD_TYPE` does not match the profile `build_type` you get errors that are tricky to diagnose as they only manifest at build time (as in the example project attached in the thread)

I don't really see a disadvantage to **not** adding the build type to the conan install commands, as if the conan `build_type` does not match `CMAKE_BUILD_TYPE`, the config files generated by `CMakeDeps` will not work.

memsharded:
> If the configurations were significantly different I would agree that it would be better to have completely different profiles for each configuration, but the scenario here is that the configurations are the same in every respect apart from the build_type, so your suggestion would lead to a proliferation of profile files that differ only by build_type.

Not necessarily. You can have a ``mydebug`` profile with just ``build_type=Debug`` setting there, and use profile composition. as the variables accept list of profiles so with ``CONAN_HOST_PROFILE="myprofile;mydebug"``, it will manage it, without having to duplicate all profiles.


> I don't really see a disadvantage to not adding the build type to the conan install commands, as if the conan build_type does not match CMAKE_BUILD_TYPE, the config files generated by CMakeDeps will not work.

It can work in different scenarios. What is not possible is to override user intent. If a user explicitly define their own profile, and they put a certain ``build_type=xxx`` in the profile and they decide to pass that profile to ``cmake-conan``, ``cmake-conan`` shouldn't overwrite the user value. One of the main design principles in Conan is that the user should have the control whenever possible, even if they "seem to be wrong", because there are a myriad of corner cases that the users need that control. In this case ``cmake-conan`` overwriting the user profile value doesn't seem the right approach.


> So if the CMAKE_BUILD_TYPE does not match the profile build_type

There is a common use case, of users wanting to have dependencies in ``Release``, while the current ``CMAKE_BUILD_TYPE=Debug``. While this is indeed very problematic in platforms as Windows, it can work fine in Linux, and many users do this. For this to work, typically is enough to do ``conan install . -s &:build_type=Debug -s build_type=Release``.

planetmarshall:
Thanks for the detailed reply, I'll take the profile composition approach.

memsharded:
> Thanks for the detailed reply, I'll take the profile composition approach.

Thanks for understanding and thanks for your feedback!

PR #633, opened 2024-03-27
Make conanfile directory configurable

valgur:
I stumbled upon a scenario where it made sense not to place the conanfile at the project root and needed to modify the conanfile directory location in `conan_provider.cmake`. I thought it would not hurt to convert it into an easily configurable variable instead of hard-coding it in general.


valgur:
Would be nice to make it a CACHE variable and group it together with other options at the bottom of the file, but it fails when converted to a CACHE var for some reason.

PR #634, opened 2024-03-27
Include license in conan_provider.cmake

valgur:
The MIT license info should be a part of `conan_provider.cmake`, imo, since it's meant to be copied as a single file in general and the license info and attribution are not really optional. Makes it slightly more convenient to use.

### DIFF
--- a/conan_provider.cmake
+++ b/conan_provider.cmake
@@ -1,3 +1,25 @@
+# The MIT License (MIT)
+#
+# Copyright (c) 2019 JFrog
+#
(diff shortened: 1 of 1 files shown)


conan_provider.cmake
```diff
@@ -1,3 +1,25 @@
+# The MIT License (MIT)
+#
+# Copyright (c) 2019 JFrog
```
memsharded:
```suggestion
# Copyright (c) 2024 JFrog
```

memsharded (approved):

PR #640, opened 2024-05-09
Remove path from tools.build:compiler_executables

ssrobins:
Addresses https://github.com/conan-io/conan/issues/16212


conan_provider.cmake
```diff
@@ -312,14 +312,16 @@ macro(append_compiler_executables_configuration)
     set(_conan_c_compiler "")
     set(_conan_cpp_compiler "")
     if(CMAKE_C_COMPILER)
-        set(_conan_c_compiler "\"c\":\"${CMAKE_C_COMPILER}\",")
+        get_filename_component(_c_compiler_name ${CMAKE_C_COMPILER} NAME)
```
memsharded, line 315: I am afraid this could be a problem.
There are users that can set the full path to a compiler in a toolchain file or the like. It is necessary that ``compiler_executables`` will contain the full path here, so dependencies will also be able to use it, otherwise they might fail.

I think the fact that Meson doesn't work with the full path to ``cl.exe`` because it contains spaces (from https://github.com/conan-io/conan/issues/16212) is not enough for this change.

In any case, we could discuss if we wanted to reduce this for MSVC ``cl.exe`` exclusively, but not for other compilers, will ask @jcar87 for feedback. And then we probably still need a fix in ``Meson`` so it doesn't fail with path with spaces there.

puetzk: Agreed, even for MSVC it's not uncommon to have multiple copies of cl.exe (via different toolsets, different architectures, or [Side-by-side Minor Version MSVC Toolsets](https://devblogs.microsoft.com/cppblog/side-by-side-minor-version-msvc-toolsets-in-visual-studio-2019/), so the full path matters.

ssrobins:
I confirmed that with the fix from https://github.com/conan-io/conan/issues/16212, this change is rendered obsolete. And from the comments, that's good because it sounds like it wouldn't have worked anyway. Thanks as always for the great community support! Closing this.

PR #641, opened 2024-05-13
Add support for MinGW Clang

valgur:
The C++ runtime is not detected correctly for MinGW Clang and Conan assumes the compiler to be ClangCL otherwise.

Here's a CI run after the fix: https://github.com/novatel/novatel_edie/actions/runs/9062747744/job/24897263836#step:6:87

### DIFF
--- a/conan_provider.cmake
+++ b/conan_provider.cmake
@@ -171,7 +171,7 @@ function(detect_lib_cxx LIB_CXX)
         set(${LIB_CXX} "libstdc++${_CONAN_GNU_LIBSTDCXX_SUFFIX}" PARENT_SCOPE)
     elseif(CMAKE_CXX_COMPILER_ID MATCHES "AppleClang")
         set(${LIB_CXX} "libc++" PARENT_SCOPE)
-    elseif(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND NOT CMAKE_SYSTEM_NAME MATCHES "Windows")
+    elseif(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND NOT MSVC)
         # Check for libc++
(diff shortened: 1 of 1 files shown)


conan_provider.cmake
```diff
@@ -171,7 +171,7 @@ function(detect_lib_cxx LIB_CXX)
         set(${LIB_CXX} "libstdc++${_CONAN_GNU_LIBSTDCXX_SUFFIX}" PARENT_SCOPE)
     elseif(CMAKE_CXX_COMPILER_ID MATCHES "AppleClang")
         set(${LIB_CXX} "libc++" PARENT_SCOPE)
-    elseif(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND NOT CMAKE_SYSTEM_NAME MATCHES "Windows")
+    elseif(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND NOT MSVC)
```
memsharded, line 174: Is ``MSVC`` variable defined for all flavors of Clang: LLVM/Clang external to Visual Studio installation and ClangCL within Visual Studio?

valgur: For the latter, yes:
> `MSVC`: Set to true when the compiler is some version of Microsoft Visual C++ **or another compiler simulating the Visual C++ cl command-line syntax.**

For an external Clang on Windows, I can only assume that it's set to false, but I have not tested this.

memsharded: I have just checked. Even for LLVM/Clang, that doesn't use ``libcxx``, but MSVC runtime, ``MSVC`` variable is not defined.
It seems that it would be good to improve the condition somehow, so that detection doesn't run for LLVM/Clang (MSVC backend)

PR #643, opened 2024-05-22
Fix _get_msvc_ide_version for VS17.10

chrarnoldus:
Visual Studio 17.10 was just released with MSVC version 1940, which gives an error: "Unknown MSVC compiler version"

https://learn.microsoft.com/sl-si/cpp/overview/compiler-versions?view=msvc-170


conan.cmake
```diff
@@ -55,7 +55,7 @@ function(_get_msvc_ide_version result)
         set(${result} 15 PARENT_SCOPE)
     elseif(NOT MSVC_VERSION VERSION_LESS 1920 AND MSVC_VERSION VERSION_LESS 1930)
         set(${result} 16 PARENT_SCOPE)
-    elseif(NOT MSVC_VERSION VERSION_LESS 1930 AND MSVC_VERSION VERSION_LESS 1940)
+    elseif(NOT MSVC_VERSION VERSION_LESS 1930 AND MSVC_VERSION VERSION_LESS 1941)
```
puetzk:
```suggestion
    elseif(NOT MSVC_VERSION VERSION_LESS 1930 AND MSVC_VERSION VERSION_LESS 1950)
```

You should raise this all the way to 1950

Per https://devblogs.microsoft.com/cppblog/msvc-toolset-minor-version-number-14-40-in-vs-2022-v17-10/ the Visual Studio 17 v143 toolset is going to continue on throughout 14.4x.

> We will make separate announcements about MSVC compiler toolset versions beyond 14.4x (i.e., versions >= 14.50) at an appropriate time in the future.

puetzk (changes requested):

puetzk (approved):
I'm not a maintainer (so this doesn't count to merge) but LGTM - it's now the same change I was going to propose...

memsharded:
I am going to merge this, even if the CI is not passing. This cmake-conan integration for Conan 1 is no longer maintained, and CI has been updated for Conan 2.

PR #648, opened 2024-06-23
Added support of Clang/LLVM toolset of Visual Studio

wizardsd:
Made it possible to use clang-cl.exe as a compiler on Windows in Visual Studio 2022.


memsharded (commented):
Thanks for your contribution @wizardsd 

PRs with changes for Conan 1.X should go to ``develop`` branch, not master.

I think the ``elseif(NOT MSVC_VERSION VERSION_LESS 1930 AND MSVC_VERSION VERSION_LESS 1950)`` was already merge to ``develop``

Also, please take not that the Conan 1.X integration is no longer recommended. If there are risks of breaking, changes might not be merged. It doesn't seem high risk in this specific case, but we need to check.

PR #654, opened 2024-07-18
Added cstd detection, and tidied variable names

jhol:


jhol:
I don't think the test failures are my fault. Looks like there is something up with the build machine.

memsharded (commented):
Thanks for your contribution @jhol 

I think the main contribution is fine, I need to double check, but I think it won't affect the default behavior as the ``cstd`` definition in profile is being discarded by Conan 2.5. Maybe we need to bump the necessary Conan version.

I see that you are also doing some other refactors/improvements. While they look good, I think it is better to separate them to another PR and keep this one focused on the ``cstd`` thing exclusively.

jhol:
> I see that you are also doing some other refactors/improvements. While they look good, I think it is better to separate them to another PR and keep this one focused on the `cstd` thing exclusively.

Ok, I've opened another PR with tidy-ups across the board: https://github.com/conan-io/cmake-conan/pull/656 . I'll come back later with another attempt at adding cstd support when/if these are accepted.

PR #656, opened 2024-07-19
Tidy ups to variable and argument names

jhol:
This patch-set makes all variables and arguments consistent with `lower_case snake_case` style, ~with a underscore prefix applied to local variables~.

PR #664, opened 2024-08-13
Set compiler.cstd in host profile

jhol:
The `cstd` setting was added to Conan in v2.4.0. Similar to `cppstd`, it is used to indicate the selected C language standard.

~The branch of this PR is based upon the `tidy-ups` branch from #656, which has not yet been merged to `develop2`, hence why the additional patches are also visible.~

### DIFF
--- a/conan_provider.cmake
+++ b/conan_provider.cmake
@@ -20,7 +20,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-set(CONAN_MINIMUM_VERSION 2.0.5)
+set(CONAN_MINIMUM_VERSION 2.4.0)
 
 # Create a new policy scope and set the minimum required cmake version so the
 # features behind a policy setting like if(... IN_LIST ...) behaves as expected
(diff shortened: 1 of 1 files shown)


jhol:
@valgur @memsharded would it be possible to get this merged?

memsharded:
Hi @jhol 

Thanks for your PR and sorry it was not reviewed back then.

The problem with this PR is that it creates a profile that is not aligned with the default Conan profiles.
``conan profile detect`` will not define the ``cstd`` yet, at the moment it is considered an ``opt-in`` as it can change some things, effect the ``package_id``, etc, so it is a bit risky to be defined by default.

I am not fully opposed to it, but we need to carefully consider this, I'll try to review it with @jcar87 when possible, thanks for the ping!

jhol:
> The problem with this PR is that it creates a profile that is not aligned with the default Conan profiles.
conan profile detect will not define the cstd yet, at the moment it is considered an opt-in as it can change some things, effect the package_id, etc, so it is a bit risky to be defined by default.

Is there a way we can make this opt-in or opt-out in `conan_provider.cmake`? This is necessary in my project so that the correct `cstd` can be passed through to conan recipes for cmake-powered projects.

PR #668, opened 2024-08-27
support list in conan command during version check

oleksandr-prykhodko-quickbase:
We have custom wrapper around `conan` command and use it in cmake as `set(CONAN_COMMAND "py;conan_command.py")`
Passing it to `conan_get_version` unwraps it as a list thus version check calls whatever is first in the list instead of the whole command.


conan_provider.cmake
```diff
@@ -560,7 +560,7 @@ macro(conan_provide_dependency method package_name)
     get_property(_conan_install_success GLOBAL PROPERTY CONAN_INSTALL_SUCCESS)
     if(NOT _conan_install_success)
         find_program(CONAN_COMMAND "conan" REQUIRED)
-        conan_get_version(${CONAN_COMMAND} CONAN_CURRENT_VERSION)
+        conan_get_version("${CONAN_COMMAND}" CONAN_CURRENT_VERSION)
```
memsharded, line 563: This is not very clear. The ``CONAN_COMMAND`` variable is the one defined in the above ``find_program()``, not a custom defined one, but you said you defined it as ``set(CONAN_COMMAND "py;conan_command.py")``. Where did you defined it? And how it gets to this location with that value?

memsharded (commented):
Thanks for your contribution @oleksandr-prykhodko-quickbase

conan_provider.cmake
oleksandr-prykhodko-quickbase: Your script defines this variable at this point, but you don't own it—we can define it too! :)

At the beginning of `CMakeLists.txt`, we set up various configurations, including the `CONAN_COMMAND`. Because of this setup, `CONAN_COMMAND` is not being set by the `find_program()` call mentioned above (as described in the [documentation](https://cmake.org/cmake/help/latest/command/find_program.html): 'If the program is found, the result is stored in the variable, and the search will not be repeated unless the variable is cleared.').

As a result, we define `CONAN_COMMAND` before any actual work is done, meaning that when `find_program()` is invoked later, it effectively does nothing.

The `conan_provider` is used as advised with `-DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=../conan_provider.cmake`. When CMake retrieves dependencies from Conan, `conan_provide_dependency` is called. By this point, our pre-setup has already executed, and `CONAN_COMMAND` has been set.

memsharded: But that approach is a bit risky, isn't it? If there is some other Conan installed in the system, the ``find_program()`` will find it and use it, while your intention was to use your own wrapper always, is that correct?

memsharded: Ah, no sorry, you are right, if the ``find_program()`` finds that it is set, it will not even try to find it, reusing, it.
Understood, thanks for the clarification!

memsharded (approved):
I think I am fine with this change, but will assign @jcar87 to review.

Thanks for your contribution @oleksandr-prykhodko-quickbase

pgeler:
@jcar87 would you mind to check this? another option is to do not pass CONAN_COMMAND and just use it directly as other functions does.

```
diff --git a/conan_provider.cmake b/conan_provider.cmake
index 1457c37..5aa650a 100644
--- a/conan_provider.cmake
+++ b/conan_provider.cmake
@@ -513,9 +513,9 @@ function(conan_install)
 endfunction()
 
 
-function(conan_get_version conan_command conan_current_version)
+function(conan_get_version conan_current_version)
     execute_process(
-        COMMAND ${conan_command} --version
+        COMMAND ${CONAN_COMMAND} --version
         OUTPUT_VARIABLE conan_output
         RESULT_VARIABLE conan_result
         OUTPUT_STRIP_TRAILING_WHITESPACE
@@ -572,7 +572,7 @@ macro(conan_provide_dependency method package_name)
     get_property(_conan_install_success GLOBAL PROPERTY CONAN_INSTALL_SUCCESS)
     if(NOT _conan_install_success)
         find_program(CONAN_COMMAND "conan" REQUIRED)
-        conan_get_version(${CONAN_COMMAND} CONAN_CURRENT_VERSION)
+        conan_get_version(CONAN_CURRENT_VERSION)
         conan_version_check(MINIMUM ${CONAN_MINIMUM_VERSION} CURRENT ${CONAN_CURRENT_VERSION})
         message(STATUS "CMake-Conan: first find_package() found. Installing dependencies with Conan")
         if("default" IN_LIST CONAN_HOST_PROFILE OR "default" IN_LIST CONAN_BUILD_PROFILE)
```

pgeler:
@memsharded any thoughts on merging this PR or if any other solution can be implemented to handle this problem?

PR #670, opened 2024-08-28
fix cppstd value for msvc compiler

oleksandr-prykhodko-quickbase:
conan [settings](https://docs.conan.io/2/reference/config_files/settings.html#settings-yml) do not support `gnu` versions of `cppstd` values in `comliper.cppstd` for `msvc`


memsharded (commented):
Thanks for your contribution @oleksandr-prykhodko-quickbase 

Quick question: why the ``CMAKE_CXX_EXTENSIONS`` will be defined if the compiler is ``msvc``? That sounds like an error in the input in the first place, if the compiler is ``msvc`` that variable shouldn't be True, isn't it?

oleksandr-prykhodko-quickbase:
> Thanks for your contribution @oleksandr-prykhodko-quickbase
> 
> Quick question: why the `CMAKE_CXX_EXTENSIONS` will be defined if the compiler is `msvc`? That sounds like an error in the input in the first place, if the compiler is `msvc` that variable shouldn't be True, isn't it?

It is indeed grey area, in the spirit of cmake and being compiler agnostic we set it `ON` for everything (as we do compile with different compilers). I think the question is whether this is an erroneous input: it really has no impact for msvc but is it necessarily restricted to set it ON (again in compiler agnostic world)?

memsharded:
> It is indeed grey area, in the spirit of cmake and being compiler agnostic we set it ON for everything (as we do compile with different compilers). I think the question is whether this is an erroneous input: it really has no impact for msvc but is it necessarily restricted to set it ON (again in compiler agnostic world)?

Thanks for the feedback.

The approach that I have seen as recommended in different places is the one in this answer: https://stackoverflow.com/questions/48026483/setting-cmake-cxx-standard-to-various-values

```cmake
if(NOT "${CMAKE_CXX_STANDARD}")
  set(CMAKE_CXX_STANDARD 11)
endif()
```
Because ``CMakeLists.txt`` should respect compiler-specifications defined in CMake toolchain files, and this is the correct way to default something without invalidating possible definitions.

Let's ask @jcar87 about this one.

memsharded (approved):
Sorry this wasn't processed before. Merging it for next 0.20 release

PR #674, opened 2024-10-04
proposal for loading conan_cmakedeps_paths.cmake

memsharded:
After https://github.com/conan-io/conan/pull/17561

This PR does:

- Update all tests to use ``CMakeConfigDeps``, explicitly
- Fix tests that changed the logic a bit because of ``CMakeConfigDeps``
- Drop tests about find modules, no longer applicable to ``CMakeConfigDeps``
- Removed the automatic injection of ``CMakeDeps``. User conanfiles should define the generator. The warning has been there for a long time already
- Updated the readme to recommend ``CMakeConfigDeps``

This doesn't really change or break users with ``CMakeDeps`` in their conanfiles.


jcar87:
cc @jjcasmar

ottmar-zittlau:
Hi,

maybe the following question is more related to the CMakeConfigDeps-generator in general than the cmake-conan-wrapper, but I came across it while testing the `conan-provider.cmake` of this PR.

I have put all files in a repository (https://github.com/ottmar-zittlau/conan-cmake-question) in case you'd like to follow my steps.

I have created a small (interface) library (in repo: test_lib) that only contains a header file.
The `package_info` of the associated `conanfile.py` reads:
```
    def package_info(self):
        self.cpp_info.builddirs = ["lib/cmake"]
        self.cpp_info.set_property("cmake_find_mode", "none")
```
This library is consumed by an executable (in repo: main_exe) by:
```
    def requirements(self):
        self.requires("test_lib/0.0.1")
```
If I follow the conventional approach (conan install + generated presets) everything works.
However, using the cmake-wrapper with the CMakeConfigDeps-generator doesn't work.
If I run
```
cmake -B build -S . -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=conan_provider.cmake -DCONAN_HOST_PROFILE="auto-cmake;../new_cmakedeps_profile"
```
cmake complains that target `testLib::testLib` can't be found.

My build folder contains `conan_cmakedeps_paths.cmake` and this file also contains some test_lib-related paths:
```
set(CMAKE_FIND_PACKAGE_PREFER_CONFIG ON)

list(APPEND CONAN_test_lib_DIR_MULTI "/home/oz/.conan2/p/b/test_3552486bd5308/p/lib/cmake")
set(CONAN_RUNTIME_LIB_DIRS "$<$<CONFIG:None>:/home/oz/.conan2/p/b/test_3552486bd5308/p/lib>" )
# Only for VS, needs CMake>=3.27
set(CMAKE_VS_DEBUGGER_ENVIRONMENT "PATH=${CONAN_RUNTIME_LIB_DIRS};%PATH%")
list(PREPEND CMAKE_PROGRAM_PATH "/home/oz/.conan2/p/b/test_3552486bd5308/p/bin")
list(PREPEND CMAKE_LIBRARY_PATH "/home/oz/.conan2/p/b/test_3552486bd5308/p/lib")
list(PREPEND CMAKE_INCLUDE_PATH "/home/oz/.conan2/p/b/test_3552486bd5308/p/include")
# Definition of CMAKE_MODULE_PATH to be able to include(module)
list(PREPEND CMAKE_MODULE_PATH "/home/oz/.conan2/p/b/test_3552486bd5308/p/lib/cmake")
```
I'm pretty confident that cmake reads this file (I printed the variable `CONAN_test_lib_DIR_MULTI` in my cmake-file to check) - so I think the changes to `conan_provider.cmake` have an effect.

I was comparing the variables set in `conan_cmakedeps_paths.cmake` with the search paths given in [cmake config search procedure](https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure) and I somehow can understand that cmake doesn't find the library. I would have expected that the path is added to either `CMAKE_PREFIX_PATH` or that `TESTLIB_ROOT` is set. The [CMakeConfigDeps documentation](https://docs.conan.io/2/reference/tools/cmake/cmakeconfigdeps.html#generated-files) states that a `TESTLIB_DIR` variable should be defined (and I can find this variable in the cmake documentation, but don't really understand if it should be set externally) - but also this variable doesn't seem to be set (my CMakeCache.txt file contains `testLib_DIR:PATH=testLib_DIR-NOTFOUND`).

How can I make this work? Do I maybe need to add something else to my library's `package_info`?

Thanks in advance and best regards
oz

memsharded:
Hi @ottmar-zittlau 

Thanks for the feedback.

We have just released ``CMakeConfigDeps``, I had these changes pending to move this integration forward, I'll do it soon.
Thanks for the repro, I'll have a look when moving this PR forward to check what could be happening.

ottmar-zittlau:
@memsharded Thanks a lot! It's not an urgent issue for me.

memsharded:
Hi @ottmar-zittlau 

I am going to merge and close this PR.
Regarding your issue above, it would be better to create a dedicated ticket for it, some quick observations:

- ``settings = ("build_type",)`` is most often not enough. At the very least the ``os`` and ``arch`` should be there, but also the  ``compiler`` is almost always needed as input, even if for some cases it can be removed later for ``package_id``
- ``tools.cmake.cmakedeps:new=will_break_next`` is no longer necessary, just use ``CMakeConfigDeps`` directly
- Update to the latest ``conan_provider.cmake`` after this PR is merged.

PR #707, opened 2025-09-12
Add support for CONAN_INSTALL_BUILD_CONFIGURATIONS

### LINKED ISSUE #705: Make the installed build configurations overridable
In the `conan_provide_dependency()` implementation, the current logic hard-codes the build configurations that are installed:

```cmake
        get_property(_multiconfig_generator GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
        if(NOT _multiconfig_generator)
            message(STATUS "CMake-Conan: Installing single configuration ${CMAKE_BUILD_TYPE}")
            conan_install(${_host_profile_flags} ${_build_profile_flags} ${CONAN_INSTALL_ARGS} ${generator})
        else()
            message(STATUS "CMake-Conan: Installing both Debug and Release")
            conan_install(${_host_profile_flags} ${_build_profile_flags} -s build_type=Release ${CONAN_INSTALL_ARGS} ${generator})
            conan_install(${_host_profile_flags} ${_build_profile_flags} -s build_type=Debug ${CONAN_INSTALL_ARGS} ${generator})
        endif()
```

However, projects may want to have direct control over which configurations of its dependencies should be used (I have multiple consulting clients where this is the case in their production projects). For example, an organisation may want to always use Release builds for its dependencies on non-Windows because they don't care about debugging into dependencies, but they may be sensitive to the runtime performance of those dependencies.

I propose adding support for a new `CONAN_INSTALL_BUILD_CONFIGURATIONS` variable. If it is not defined, the logic is the same as right now. If the variable is defined, we install the specified dependencies instead. I have a relatively small patch with that change ready, although it still needs doc and test updates. I may need some guidance on the test parts.

craigscott-crascit:
Let the user or project override the set of build types we execute the conan install command for. This is most useful for enabling projects to build only one configuration when using a multi-config generator, but it can also be used to install for multiple build types with a single-config generator.

Fixes: #705


craigscott-crascit:
@jcar87 I've updated the docs, but don't know what you want to do for the tests. If you're happy to take care of that, please go ahead and push commits to my branch as needed.

memsharded (approved):
This is looking good, many thanks for your contribution @craigscott-crascit 

I'd like to have the feedback from @jcar87 too.

craigscott-crascit:
I fixed the failing test. Would be good to trigger another round of CI jobs to confirm that was the only problem.

craigscott-crascit:
I also added a couple of tests for the new `CONAN_INSTALL_BUILD_CONFIGURATIONS` variable. Not sure if that's following the style you're after, but hopefully it's in the ballpark.

jcar87:
Thanks @craigscott-crascit  for the PR and for taking the time to implement the tests - our test suite in `cmake-conan` is not very ergonomic, so we appreciate it!

Need to rebase this on top of `develop2` after https://github.com/conan-io/cmake-conan/pull/698 was merged, happy to do so ourselves and take over the rest if you're okay with that? (although I suspect it will pass the tests just fine)

craigscott-crascit:
>Need to rebase this on top of develop2 after https://github.com/conan-io/cmake-conan/pull/698 was merged, happy to do so ourselves and take over the rest if you're okay with that?

Yes, please take it over and push changes as you wish. Thanks.

jcar87:
The tests are failing for a very interesting set of circumstances. (The main being that in that Test class, the build directory is reused and doing an actual build highlights an actual issue with the Conan `CMakeDeps` generator) 

Need to discuss with @memsharded  tomorrow, when we have a single config CMake generator, runing `conan install` on multiple configurations will cause files to overwrite each other in the generators folder, and if the installed configuration differs from the "self" configuration, we need to use a the `&:build_pattern=${CMAKE_BUILD_TYPE}` to tell Conan what the actual build time for the top-level self project is.

For a multi config generator, I'm not sure 100% it would be an issue, but double checking. 

for what it's worth, all of these problems should go away with the `CMakeConfigDeps` generators, which intends to output files that much closer to what CMake itself would (with the added bonus of being able consume config files inside the packages too)

conan_provider.cmake
```diff
@@ -591,24 +591,37 @@ macro(conan_provide_dependency method package_name)
...
+            
         endif()
         list(JOIN _build_configs ", " _build_configs_msg)
         message(STATUS "CMake-Conan: Installing configuration(s): ${_build_configs_msg}")
         foreach(_build_config IN LISTS _build_configs)
-            conan_install(${_host_profile_flags} ${_build_profile_flags} -s build_type=${_build_config} ${CONAN_INSTALL_ARGS} ${generator})
+            set(_self_build_config "")
+            if(_multiconfig_generator AND NOT _build_config STREQUAL "${CMAKE_BUILD_TYPE}")
+                set(_self_build_config "-s &:build_type=${CMAKE_BUILD_TYPE}")
+            endif()
```
craigscott-crascit: `CMAKE_BUILD_TYPE` should not be set at all if we're using a multi-config generator. What is this condition trying to do?

conan_provider.cmake
```diff
@@ -591,24 +591,37 @@ macro(conan_provide_dependency method package_name)
...
         endif()
+
+        get_property(_multiconfig_generator GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
+        
         if(DEFINED CONAN_INSTALL_BUILD_CONFIGURATIONS)
             # Configurations are specified by the project or user
             set(_build_configs "${CONAN_INSTALL_BUILD_CONFIGURATIONS}")
+            list(LENGTH _build_configs _build_configs_length)
+            if(NOT _multiconfig_generator AND _build_configs_length GREATER 1)
+                message(FATAL_ERROR "cmake-conan: please only specify a single configuration in CONAN_INSTALL_BUILD_CONFIGURATIONS")
```
craigscott-crascit: If I understand correctly, this restriction only applies when using a single-config generator. The error message should mention that somewhere to avoid potential confusion.

README.md
```diff
@@ -68,10 +68,25 @@ If you need to customize the profile, you can do so by modifying the value of `C
...
+Any `build_type` specified by a host profile is ignored.
+The `build_type` is handled separately, and the default behavior depends on the CMake generator used:
+* For single-configuration generators, `conan install` is executed once with `build_type` set to the `CMAKE_BUILD_TYPE`. The `CMAKE_BUILD_TYPE` must not be empty.
+* For multi-configuration generators, `conan install` is executed twice: once with `build_type` set to `Release`, and a second time with `build_type` set to `Debug`.
+
+You can override the default build type(s) by setting the `CONAN_INSTALL_BUILD_CONFIGURATIONS` CMake variable to a list of build types.
+`conan install` will then be invoked once for each build type.
+This can be used with both single- and multi-configuration generators.
+For example:
+* `-DCONAN_INSTALL_BUILD_CONFIGURATIONS=Release;Debug`: execute `conan install` for both `Release` and `Debug` build types, even for single-configuration generators.
```
craigscott-crascit: If we're saying only a single value can be given in `CONAN_INSTALL_BUILD_CONFIGURATIONS` when using a single-config generator, then this needs to be reworded.

conan_provider.cmake
jcar87: You're absolutely right - the condition is when we're not in a multi-config generator.

conan_provider.cmake
jcar87: Indeed - added a clarification

README.md
jcar87: Agreed, working on this!

conan_provider.cmake
```diff
@@ -599,7 +599,8 @@ macro(conan_provide_dependency method package_name)
             set(_build_configs "${CONAN_INSTALL_BUILD_CONFIGURATIONS}")
             list(LENGTH _build_configs _build_configs_length)
             if(NOT _multiconfig_generator AND _build_configs_length GREATER 1)
-                message(FATAL_ERROR "cmake-conan: please only specify a single configuration in CONAN_INSTALL_BUILD_CONFIGURATIONS")
+                message(FATAL_ERROR "cmake-conan: when using a single-config CMake generation, "
```
craigscott-crascit:
```suggestion
                message(FATAL_ERROR "cmake-conan: when using a single-config CMake generator, "
```

craigscott-crascit:
> ...all of these problems should go away with the `CMakeConfigDeps` generators, which intends to output files that much closer to what CMake itself would (with the added bonus of being able consume config files inside the packages too)

I cannot express how happy the end of that comment makes me. :)

README.md
craigscott-crascit: I pushed a commit to update the docs to mention the restriction for single-config generators.

craigscott-crascit:
Is there anything left to do for this now? If so, let me know what we're waiting on in case it's something I can address. I'd like to get this PR merged soon if we can.

jcar87:
Hi @craigscott-crascit - apologies for the delay, this found me during conference season :(

I think the PR is good to be merged - the only comment I had pending publish was regarding the documentation:

> Any `build_type` specified by a host profile is ignored.
> The `build_type` is handled separately, and the default behavior depends on the CMake generator used:

That first sentence had me question everything 😅 

As it turns out, the `build_type` setting is not _always_ ignored when using single config CMake generators. 
If the user removes `auto-cmake` from `CONAN_HOST_PROFILE`, or performs profile composition by setting that variable to something like `default;auto-cmake;other-profile` - where `other-profile` (re)defines the `build_type`, then that will take priority regardless. ~Although I suspect it wasn't properly to consume dependencies until changes in this PR (the `-s &:build_type=...`).~  Although if the build_type was different than `CMAKE_BUILD_TYPE`, it would've only worked if the users had the `&:build_type` AND `build_type` in their profiles.

For multi-config generators, it was unconditionally true: we were very explicitly installing Debug and Release, irrespective of any other variable (profile, `CMAKE_BUILD_TYPE` or otherwise).

~This does leave us with two different ways of overriding the build type for dependencies when using a single config generator  but I don't think it's too problematic. I'll reword that bit of the proposed documentation and merge this.~ Apologies it took long. 

Edit: actually, this PR _is_ unconditionally overriding the build_type to obey either `CMAKE_BUILD_TYPE` or `CONAN_INSTALL_BUILD_CONFIGURATIONS` where it previously wasn't the case for single-config. 

Not sure how to proceed as all options seem valid to me:
- leave the PR as is, the documentation matches the behaviour changes. However, it may result in a behaviour changes for any users that previously did *not* include `auto-cmake` in `CONAN_HOST_PROFILE`, it can turn some previously working cases into not working anymore
- make it such that if `CONAN_INSTALL_BUILD_CONFIGURATIONS` is provided for a single-config generator build, that that value is instead baked into the "autodetected" profile instead of `CMAKE_BUILD_TYPE` -> this is the most behaviour preserving change, however this has a caveat: if `auto-cmake` is not part of `CONAN_HOST_PROFILE`, then `CONAN_INSTALL_BUILD_CONFIGURATIONS` will be completely ignored, and I'm concerned it can create confusion with regards to the expectations (we can also make it an error to have both, if that helps)

What do you think?

craigscott-crascit:
>* make it such that if `CONAN_INSTALL_BUILD_CONFIGURATIONS` is provided for a single-config generator build, that that value is instead baked into the "autodetected" profile instead of `CMAKE_BUILD_TYPE` -> this is the most behaviour preserving change, however this has a caveat: if `auto-cmake` is not part of `CONAN_HOST_PROFILE`, then `CONAN_INSTALL_BUILD_CONFIGURATIONS` will be completely ignored, and I'm concerned it can create confusion with regards to the expectations (we can also make it an error to have both, if that helps)

I'd be reluctant to go down this route. It seems too fragile to me, and the use case you highlighted does seem likely to be not all that rare. At least one of the projects I'm involved in that motivated this change would potentially hit this scenario.

>Edit: actually, this PR is unconditionally overriding the build_type to obey either `CMAKE_BUILD_TYPE` or `CONAN_INSTALL_BUILD_CONFIGURATIONS` where it previously wasn't the case for single-config.
>...
>* leave the PR as is, the documentation matches the behaviour changes. However, it may result in a behaviour changes for any users that previously did not include `auto-cmake` in `CONAN_HOST_PROFILE`, it can turn some previously working cases into not working anymore

If users were specifying a `CONAN_HOST_PROFILE` and did not include `auto-cmake` (or they appended a profile that changed the `build_type` after `auto-cmake`), and if that gave a different result to what they'd get with this PR's changes, then it could be argued that setup is likely to be wrong. As you mentioned, they would have to do the `&:build_type` AND `*:build_type` dance for that to be correct.

I'm leaning towards the "leave the PR as is" option, since that should do the right thing out of the box for the vast majority of users. If the change in behavior breaks those who run afoul of the above scenario, they can pin their provider to the existing cmake-conan commit and deal with the change at their convenience (they should be doing that anyway instead of tracking a floating branch head). We might want to document the behavior change or highlight the scenario somewhere though. I defer to you on how/where to do that.

jcar87:
> If users were specifying a CONAN_HOST_PROFILE and did not include auto-cmake (or they appended a profile that changed the build_type after auto-cmake), and if that gave a different result to what they'd get with this PR's changes, then it could be argued that setup is likely to be wrong. As you mentioned, they would have to do the &:build_type AND *:build_type dance for that to be correct.
> 
> I'm leaning towards the "leave the PR as is" option

Thanks so much for the feedback! Happy to leave it as is, you are right, it may be a behaviour change but at the same time, the bit of it that "worked" was rather fragile so I'm inclined to believe it wasnt widely used (if at all) - let's wait and see :)

jcar87 (approved):
Thanks for the work and your patience!

Issue #706, opened 2025-08-18
Make it possible to capture the JSON output of the conan install command(s)


craigscott-crascit:
In the implementation of the `conan_install()` function, the `conan install` command output is specified to be in JSON format and is captured to a CMake variable. However, the contents of that variable are not made available to the project in any way. One of the really handy uses of that output is to check which packages ended up being built from source instead of retrieving a pre-built binary package from one of the remotes. It can also be used as part of uploading just-built binary packages when one or more dependencies had to be built-on-demand due to missing binary packages.

One potential complication is where multiple calls to `conan_install()` are made. Right now, that's the case when a multi-config generator is used. #705 proposed a feature where the project can specify whichever build configuration(s) it wants, including supporting multiple configurations for a single-config generator if they want to. So a feature like what is being proposed here would need some way of saving the results of multiple runs and the project being able to differentiate the separate per-config output.

Perhaps one solution would be to save the output to config-specific files in the output folder. That wouldn't require any new keyword or variable, we could just write these files always. That would certainly help with debugging what's going on with fulfilling dependencies. I use such a file frequently in production projects where I can't currently use `cmake-conan`, and it would solve some problems I face with those projects that do use `cmake-conan`.

jcar87:
Hi @craigscott-crascit - thanks for opening this issue.

If I'm reading this correctly, I think it would be sensible just to output these files somewhere to `CMAKE_BINARY_DIR`, and perhaps choose different filenames in case there are multiple invocations, right? (on the later point, I think they could be merged into a single one with `conan pkglist` ([docs](https://docs.conan.io/2/reference/commands/pkglist.html)), but I'd need to check with team to make sure that is the case)

craigscott-crascit:
Merging them when there's more than one build type/configuration might simplify things for the project, especially if combined with my suggestion in #705 for allowing the project to specify the build type(s) they want installed. I'm already doing the merge packagelist thing in at least one project where I've been carrying patches to cmake-conan for this.

As for the directory where we write the file(s), why not put them in the conan generation directory along with all the other files conan writes?

jcar87:
> Merging them when there's more than one build type/configuration might simplify things for the project, especially if combined with my suggestion in https://github.com/conan-io/cmake-conan/issues/705 for allowing the project to specify the build type(s) they want installed. I'm already doing the merge packagelist thing in at least one project where I've been carrying patches to cmake-conan for this.

[...]

PR #708, opened 2025-09-12
Remove malformed cmake_parse_arguments() call that has no effect

### LINKED ISSUE #704: Malformed cmake_parse_arguments() at start of conan_install() function
The start of the `conan_install()` function looks like this:

```cmake
    cmake_parse_arguments(ARGS conan_args ${ARGN})
    set(conan_output_folder ${CMAKE_BINARY_DIR}/conan)
    set(conan_args ${conan_args} -of=${conan_output_folder})
```

That call to `cmake_parse_arguments()` is not a properly formed call. The syntax of that call from the CMake docs looks like this:

```cmake
cmake_parse_arguments(<prefix> <options> <one_value_keywords>
                      <multi_value_keywords> <args>...)
```

I'm not sure what the original intent of the call was, but it is currently almost certain to be doing nothing. That is more by good luck than anything. The `one_value_keywords` and `multi_value_keywords` are going to be taken from the first arguments of `${ARGN}`, and I'm pretty sure that was never the intention. The variables defined by the call will be named `ARGS_...`, and I don't see any code that makes use of any such variables. The `conan_args` is going to be treated as an option keyword when parsing arguments, it is not going to result in a variable named `conan_args` being set. I think this call to `cmake_parse_arguments()` should probably be deleted.

The `conan_args` variable is used in later code where it is always followed by `${ARGN}` and used in a `conan install` command line. Perhaps the best way to clean this up would be to define `conan_args` like so:

```cmake
    set(conan_args -of=${conan_output_folder} ${ARGN})
```

Then in the two places later on where `${conan_args}` is used, we don't need to put `${ARGN}` after it.

Also note that the way `${ARGN}` is used, it will discard any empty items in that list. It probably hasn't mattered so far for the arguments we're supporting here, but for this to be more robust, we'd need to be much more careful about how we assign, quote, and use it and any variables derived from it. My [Quoting In CMake](https://crascit.com/2022/01/25/quoting-in-cmake/) blog post goes into the details if you want to explore that further, especially the [Passing Lists As Command Arguments](https://crascit.com/2022/01/25/quoting-in-cmake/#h-passing-lists-as-command-arguments) section.

craigscott-crascit:
Fixes: #704

### DIFF
--- a/conan_provider.cmake
+++ b/conan_provider.cmake
@@ -461,7 +461,6 @@ endfunction()
 
 
 function(conan_install)
-    cmake_parse_arguments(ARGS conan_args ${ARGN})
     set(conan_output_folder ${CMAKE_BINARY_DIR}/conan)
     # Invoke "conan install" with the provided arguments
     set(conan_args ${conan_args} -of=${conan_output_folder})


craigscott-crascit:
The description of #704 suggests we could update `conan_args` like so:

```cmake
set(conan_args -of=${conan_output_folder} ${ARGN})
```

I've decided not to do that in this PR, as I will include that change as part of the fix for #706. I need to wait for #707 to be merged before I can put up that change though, so I'm getting this part out of the way now as a smaller, more focused PR.

memsharded (commented):
Thanks very much for your contribution @craigscott-crascit 

I see what is the issue, and I agree with your issue report that it probably would be better to clean up the code a bit further to clarify the intention, as the ``set(conan_args ${conan_args} -of=${conan_output_folder})`` has no sense if ``${conan_args]`` is empty. And it would be better to clearly define ``conan_args`` with the ``-of`` + ``$ARGN`` as you suggested.

What do you think?

craigscott-crascit:
I'm going to change that line as soon as #707 is merged. I want to add the `build_type` as a function argument, then include that in `conan_args`. The code I have waiting looks something like this:

```cmake
function(conan_install build_type)   # <--- note extra function argument added
    set(conan_output_folder ${CMAKE_BINARY_DIR}/conan)
    set(conan_args -of=${CONAN_OUTPUT_FOLDER})
    # New code from here down
    if(build_type STREQUAL "")
        # We can't assume the build type will be ${CMAKE_BUILD_TYPE} because it may be specified
        # as something else by a profile provided by the project or user through CONAN_INSTALL_ARGS.
        set(conan_stdout_file ${CONAN_OUTPUT_FOLDER}/conan_install_log.json)
    else()
        set(conan_stdout_file ${CONAN_OUTPUT_FOLDER}/conan_install_log-${build_type}.json)
        list(APPEND conan_args -s build_type=${build_type})
    endif()
```

Then later on, I write out the contents of the `${conan_stdout}` variable to the file at `${conan_stdout_file}`. The `${ARGN}` has to come after the `-s build_type=${build_type}` part, so I wanted to hold off appending it to `conan_args` until I do the above. That will reduce the churn in the git diffs around this area.

EDIT: I will probably end up with a simpler version of the above, since we will be able to guarantee `build_type` won't be empty. But I've copy-and-pasted the above from an earlier version I had when working on this a few months ago. The final version will be simpler, but there will need to be some decisions made as part of addressing #706.

jcar87:
LGTM! Rebased on top of most recent `develop2` to sync up with fixed github actions

jcar87 (approved):

PR #712, opened 2025-11-05
Use CMAKE_TOOLCHAIN_FILE if provided

JulianH1989:
I have tested this wrapper on cross arm_none_eabi builds with windows.
Turns out it failed because conan was ignoring my toolchain file and letting cmake guess compiler arguments which made my cross arm none eabi build fail.

=> Test if the user has set CMAKE_TOOLCHAIN_FILE via CLI or via preset. If this is set pass this to conan

### DIFF
--- a/conan_provider.cmake
+++ b/conan_provider.cmake
@@ -425,6 +425,17 @@ function(detect_host_profile output_file)
     string(APPEND profile "[conf]\n")
     string(APPEND profile "tools.cmake.cmaketoolchain:generator=${CMAKE_GENERATOR}\n")
 
+    # if a custom cmake toolchain was specified use it
+    if(CMAKE_TOOLCHAIN_FILE AND NOT EXISTS ${CMAKE_TOOLCHAIN_FILE})
+      message(FATAL_ERROR "Specified toolchian file ${CMAKE_TOOLCHAIN_FILE} does not exist")
+    endif()
+
+    if(CMAKE_TOOLCHAIN_FILE AND EXISTS ${CMAKE_TOOLCHAIN_FILE})
+      # pass cmake toolchain file to conan
+      string(APPEND profile "tools.cmake.cmaketoolchain:toolchain_file=${CMAKE_TOOLCHAIN_FILE}\n")
+  endif()
+
+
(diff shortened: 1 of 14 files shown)


conan_provider.cmake
```diff
@@ -425,6 +425,17 @@ function(detect_host_profile output_file)
     string(APPEND profile "[conf]\n")
     string(APPEND profile "tools.cmake.cmaketoolchain:generator=${CMAKE_GENERATOR}\n")
 
+    # if a custom cmake toolchain was specified use it
+    if(CMAKE_TOOLCHAIN_FILE AND NOT EXISTS ${CMAKE_TOOLCHAIN_FILE})
+      message(FATAL_ERROR "Specified toolchian file ${CMAKE_TOOLCHAIN_FILE} does not exist")
+    endif()
```
memsharded: Is this really necessary? I mean, if the toolchain file is not there when cmake is invoked, cmake will already error, so it will never get to this point in the ``conan_provider.cmake``?

memsharded (commented):
I think it looks good, it would be nice to have this covered by one test. Please let us know if you could do it, otherwise we could try to help.

Please note that providing a CMake toolchain file and propagating it to dependencies via the Conan ``conf`` will not be enough for dependencies that use other build systems, and there are quite a few of them in ConanCenter. To be able to have dependencies built correctly for those cases, it might still be necessary to provide the necessary information in Conan profile files.

I am passing this to @jcar87 for review.

conan_provider.cmake
JulianH1989: I guess you are right. Removed that double check

JulianH1989:
@memsharded  I added a test repo. However i'm not familiar with Github Workflows.
Could you add the test project to the workflows.

invoke it with
```
cmake --preset arm_none_eabi_gcc_14_win
cmake --build --preset cmake --preset arm_none_eabi_gcc_14_win_rel
```

and it should work. i have tested it locally on my windows 11 machine.
There is a linux preset as well which should work

JulianH1989:
hey guys, any further comments on this :) ?

PR #719, opened 2025-12-03
Save JSON and stderr output from conan install to files

### LINKED ISSUE #706: Make it possible to capture the JSON output of the conan install command(s)

craigscott-crascit:
This allows the user or project to do further processing of the JSON output (e.g. to publish missing binary packages that were built) and to troubleshoot any problems during the install.

Fixes: #706

### DIFF
--- a/conan_provider.cmake
+++ b/conan_provider.cmake
@@ -35,6 +35,10 @@ set(CONAN_MINIMUM_VERSION 2.0.5)
 cmake_policy(PUSH)
 cmake_minimum_required(VERSION 3.24)
 
+# Handling of empty values for single-value keywords with cmake_parse_arguments()
+if(POLICY CMP0174)
+    cmake_policy(SET CMP0174 NEW)
+endif()
 
 function(detect_os os os_api_level os_sdk os_subsystem os_version)
     # it could be cross compilation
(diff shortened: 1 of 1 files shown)


conan_provider.cmake
```diff
@@ -35,6 +35,10 @@ set(CONAN_MINIMUM_VERSION 2.0.5)
 cmake_policy(PUSH)
 cmake_minimum_required(VERSION 3.24)
 
+# Handling of empty values for single-value keywords with cmake_parse_arguments()
+if(POLICY CMP0174)
+    cmake_policy(SET CMP0174 NEW)
+endif()
```
craigscott-crascit, line 41: Instead of this, we could update the `cmake_minimum_required()` call above to specify a version range that included up to at least 3.31. But that's a more involved change because it would require checking all policies added after CMake 3.24 up to whatever upper limit was specified. That's a separate focus to this PR, but would be good to do in a follow-up PR.

craigscott-crascit:
I wasn't sure if this needed any documentation changes. It's not currently documented, but if we want it to be officially supported and maintain the location, we could document it. Or we could leave it undocumented initially to give folks who are interested time to try it. If we don't get any negative feedback after 6 months or so, we could make it officially documented then.

craigscott-crascit:
@jcar87 Just checking if there's anything more you need me to do for this before it can be merged?

PR #723, opened 2026-02-04
feat: Auto-build packages from Git recipes when not found in remotes

tjwrona:
## Summary

**Ever been blocked by a missing package?** Your private library isn't on ConanCenter. Your colleague's experimental branch isn't published yet. That patched fork you need isn't cached anywhere. **Build grinds to a halt.**

**Not anymore.**

This PR introduces **automatic recipe-from-source building** — cmake-conan now seamlessly fetches and builds packages directly from Git repositories when they can't be found in any configured remote or local cache. Just annotate your requirements with a `#recipe:` comment, and cmake-conan handles the rest.

**conanfile.py:**
```python
def requirements(self):
    self.requires("mypackage/1.2.3")  #recipe: https://github.com/user/mypackage.git
    self.requires("private/2.0.0@myuser/stable")  #recipe: https://github.com/org/private.git
```

**conanfile.txt:**
```ini
[requires]
mypackage/1.2.3  #recipe: https://github.com/user/mypackage.git
private/2.0.0@myuser/stable  #recipe: https://github.com/org/private.git
```

**Zero friction. Zero manual intervention. Just works.**

## The Problem This Solves

Teams using Conan face a common pain point: **not every package lives on ConanCenter or a configured remote**. This creates friction in several scenarios:

- 🔒 **Private/internal packages** that can't be published publicly
- 🧪 **Experimental branches** during active development
- 🔧 **Patched forks** of upstream libraries with custom fixes
- 🚧 **New packages** not yet submitted to ConanCenter
- 🏢 **Air-gapped environments** where setting up a full Artifactory server is overkill

Previously, developers had to manually run `conan create` before their CMake build, breaking the seamless dependency provider experience. **This PR eliminates that manual step entirely.**

## How It Works

When cmake-conan encounters a `#recipe:` annotation, it performs intelligent caching:

| Scenario | Action |
|----------|--------|
| Package exists in remote | ✅ Use it directly (no build) |
| Package in cache + version is a Git tag | ✅ Reuse cached (tags are immutable) |
| Package in cache + version is a Git branch | 🔄 Compare commits, rebuild only if changed |
| Package not found anywhere | 🔨 Clone repo → `conan create` → continue |

### Smart Version Resolution

Versions are automatically mapped to Git refs:

| Conan Version | Git Ref |
|---------------|---------|
| `1.2.3` | `v1.2.3` (tag) |
| `develop` | `develop` (branch) |

**Tags = Reproducible builds.** Once built, never rebuilt unnecessarily.

**Branches = Development flexibility.** Automatically rebuilds when upstream changes.


memsharded:
Hi @tjwrona 

Thanks for your contribution.

Unfortunately, it will not be possible to move this forward.

The design of decoupling the dependencies references (like ``pkg/version``) from the origin/server (either a Conan remote or a git repo) which is abstracted away in recipes is fully intentional, and we plan to keep it that way. Both for Conan itself but also for ``cmake-conan``. 

There are many reasons for that, for example reproducibility using lockfiles, which is broken by this approach as "recipe-revisions" cannot exist in git repos. This is the same limitation that exists in ``local-recipes-index`` feature, and this is why this is not recommended to be used as a regular resolution mechanism, but as a helper to create your own binaries from a ConanCenter-like repo and put them in a proper remote server.

The ``cmake-conan`` integration is focused on one responsibility, calling ``conan install`` to fetch the dependencies from remotes when a ``find_package()`` is found when executing from CMake. This resolution of dependencies by getting the Git urls and cloning is too much, and not following the general Conan design and functionality.

The good news is that using your own ``conan_provider.cmake`` is very simple, at the end of the day, the provider is just a CMake file that lives in the repo, so if you find the functionality convenient, it will be very easy for you to use it. But I am afraid that we cannot take ownership of this functionality and its future maintenance, it adds quite a complexity for a value/feature not aligned with the general Conan design.

Thanks very much for your contribution anyway!

tjwrona:
@memsharded thank you for the quick response!

I understand, and your reasoning is definitely valid based on the core design principles Conan is trying to follow.

Funnily enough, I actually had not heard of the `local-recipe-index` feature and it looks like I basically just recreated it. 😂 As far as I can tell my approach is actually a little simpler to use than `local-recipe-index` though. (This feature automates the Git checkout directly as part of the CMake generation process and doesn't require manually cloning repos into your local recipe index to consume them)

That said, I do have a follow-up question about the reproducibility concern:

> If I understand correctly, wouldn't `--build=missing` on any recipe that uses `FetchContent` or unpinned downloads during the build have the same issue? When `conan install` goes to rebuild it from source it would run CMake generation and it is very common practice for developers to use `FetchContent` to grab code directly from Git repositories. From what I understand, lockfiles wouldn't be able to capture what those external fetches resolve to either.

I assume ConanCenter probably enforces policies to prevent this? But for private recipes or packages not on ConanCenter, it seems like the same class of problem would already exist. If that's the case, my feature wouldn't introduce a *new* failure mode, it would just makes an existing edge case more visible.

I'm not trying to argue for inclusion, I respect the decision! I'm just trying to fully understand the boundaries so I can make informed tradeoffs in my fork.

With that in mind: do you have any additional wisdom about potential gotchas I might run into? Or ways I could enhance the design to mitigate the reproducibility concerns?

memsharded:
> Funnily enough, I actually had not heard of the local-recipe-index feature and it looks like I basically just recreated it. 😂 As far as I can tell my approach is actually a little simpler to use than local-recipe-index though. (This feature automates the Git checkout directly as part of the CMake generation process and doesn't require manually cloning repos into your local recipe index to consume them)

Yes, this is something that we also explicitly excluded from "local-recipes-index", note from https://docs.conan.io/2/devops/devops_local_recipes_index.html

> The local-recipes-index repositories point to local folders in the filesystem. While users may choose to sync that folder with a git repository or other version control mechanisms, Conan is agnostic to this, as it is only aware of the folder in the filesystem that points to the (current) state of the repository. Users may choose to run git commands directly to switch branches/commit/tags and Conan will automatically recognise the changes

So the responsibility of cloning the "local-recipes-index" repo and checking out the commit, branch, etc. belongs to the user, not Conan, because that would also create some reproducibility expectations that will not be fully supported, like the last paragraph in that docs page says:

> Furthermore, this feature does not support placing server URLs directly in recipes; remote repositories must be explicitly added with conan remote add. Decoupling abstract package requirements, such as “zlib/1.3.1”, from their specific origins is crucial to resolving dependencies correctly and leveraging Conan’s graph capabilities, including version conflict detection and resolution, version-ranges resolution, [opting into pre-releases](https://docs.conan.io/2/devops/versioning/resolve_prereleases.html#resolve-prereleases-summary), [platform_requires](https://docs.conan.io/2/reference/config_files/profiles.html#reference-config-files-profiles-platform-requires), [replace_requires](https://docs.conan.io/2/reference/config_files/profiles.html#reference-config-files-profiles-replace-requires), etc. This separation also facilitates the implementation of modern DevOps practices, such as package immutability, full relocatability and package promotions.

Regarding: 

> I assume ConanCenter probably enforces policies to prevent this? But for private recipes or packages not on ConanCenter, it seems like the same class of problem would already exist. If that's the case, my feature wouldn't introduce a new failure mode, it would just makes an existing edge case more visible.

Yes, ConanCenter recipes cannot "vendor" transitive dependencies. For a package recipe to be accepted, they should fully unvendor any possible dependencies (in the form of private clones, "third_party" folders with copies, submodules, FetchContent, etc) and convert them to proper Conan packages, then use ``requires()`` properly to model those dependencies. Otherwise there would be constantly linker errors because of duplicated symbols or ODR violations.

This is why an abstraction, which is the "pkgname/version" reference that models a dependency is used.

The problem with your feature is that it goes down the classic rabbit hole. It looks relatively innocent on a pure "consumer" side recipe. But edge and rough edges can start to appear quickly. What happens if a given conanfile defines a:
```python
self.requires("zlib/1.3.1") # https://some/url/zlib.git
```
And then another conanfile in the same machine defines
```python
self.requires("zlib/1.3.1") # https://other/url/zlib.git
```
on a different URL? What would you do? It is not possible to guarantee they are the same, even if the tag the correct version, unless fully cloning checking out and comparing bit by bit, and that could make the system slow in practice. And if they don't match, then the second one will be rejected? Or just continue because zlib/1.3.1 was found in the cache? Surely the user with the second one will be annoyed in either case, because blocked or even worse, using a different source that the one they pointed in their conanfile git url!
Then, suddenly,  the system would need a record of all installed packages from a url, saving that url, then check when those are requested, to compare the previous URLs...

And that would be quickly also requested as a regular way for regular dependencies too, so any package in the graph could resolve to other one by fetching the git repo. But that would have a lot of problems, not only the highlighted above, but also the resolution of that could make things way slower, because something like an ``-update`` to update to the latest versions that is resolved with fast API calls, would suddenly require either doing clones, or even if they are cached, special git tricks to get updated information from the git remotes looking for the newer tags... too complex.


Note the git-url approach has been implemented in package managers like pip for Python or NPM, where the installation of packages is mostly "flat" (you cannot install and depend on 2 different versions of the same thing in the same place, like Python environments that are completely flat, or NPM that is flat or nested), but there are other package managers like Maven (which Conan also takes some inspiration from) that also haven't implement any git-url package resolution at all. The challenges, limitations and issues are definitely not worth it, and trying to do that can have a huge toll on the project future maintenance, development and support, without really making users much happier due to the many possible rough edges.

PR #727, opened 2026-04-28
Support CMAKE_MAP_IMPORTED_CONFIG_<CONFIG> in detect_build_type

simfinite:
Essentially, as title says. 

In CMake, [CMAKE_MAP_IMPORTED_CONFIG_<CONFIG>](https://cmake.org/cmake/help/latest/variable/CMAKE_MAP_IMPORTED_CONFIG_CONFIG.html) may be used to map custom build types (e.g. coverage) to build types/configurations known to dependencies. This could be supported as suggested in the change to the `detect_build_type` function.

Thank you for considering this.

### DIFF
--- a/conan_provider.cmake
+++ b/conan_provider.cmake
@@ -312,6 +312,15 @@ function(detect_build_type build_type)
     if(NOT multiconfig_generator)
         # Only set when we know we are in a single-configuration generator
         # Note: we may want to fail early if `CMAKE_BUILD_TYPE` is not defined
+        # Also: If CMAKE_MAP_IMPORTED_CONFIG_<CONFIG> is set, use its value
+        # as the Conan build type instead of the current CMake build type, see also:
+        # https://cmake.org/cmake/help/latest/prop_tgt/MAP_IMPORTED_CONFIG_CONFIG.html#map-imported-config-config
+        string(TOUPPER "${CMAKE_BUILD_TYPE}" CMAKE_BUILD_TYPE_UPPER)
+        if(CMAKE_MAP_IMPORTED_CONFIG_${CMAKE_BUILD_TYPE_UPPER})
+            set(${BUILD_TYPE} ${CMAKE_MAP_IMPORTED_CONFIG_${CMAKE_BUILD_TYPE_UPPER}} PARENT_SCOPE)
+        else()
+            set(${BUILD_TYPE} ${CMAKE_BUILD_TYPE} PARENT_SCOPE)
+        endif()
         set(${build_type} ${CMAKE_BUILD_TYPE} PARENT_SCOPE)
     endif()
 endfunction()
--- a/conan_provider.cmake
+++ b/conan_provider.cmake
@@ -620,9 +620,15 @@ macro(conan_provide_dependency method package_name)
             if(_multiconfig_generator)
                 set(_build_configs Release Debug)
             else()
-                set(_build_configs ${CMAKE_BUILD_TYPE})
+                # Check if Conan dependencies should be build with a different build type
+                # defined by variable CMAKE_MAP_IMPORTED_CONFIG_<CMAKE_BUILD_TYPE>
+                string(TOUPPER "${CMAKE_BUILD_TYPE}" CMAKE_BUILD_TYPE_UPPER)
+                if(CMAKE_MAP_IMPORTED_CONFIG_${CMAKE_BUILD_TYPE_UPPER})
+                    set(_build_configs ${CMAKE_MAP_IMPORTED_CONFIG_${CMAKE_BUILD_TYPE_UPPER}})
+                else()
+                    set(_build_configs ${CMAKE_BUILD_TYPE})
+                endif()
             endif()
-            
         endif()
 
         list(JOIN _build_configs ", " _build_configs_msg)


memsharded (commented):
Hi @simfinite 

Thanks for your contribution
But I am afraid this doesn't look correct:

- The value of CMAKE_MAP_IMPORTED_CONFIG can be a list of values, not just a single value.
- The fact that this variable is defined shouldn't enforce to use that configuration. This is defined as a fallback in case the defined configuration is not there. But this definition automatically changes the "detected" build-type if the variable is defined
- There is still a ``set(${build_type} ${CMAKE_BUILD_TYPE} PARENT_SCOPE)`` that defines the outside value, overwriting previous value in the ``if-else``

Something like this would need a more elaborated rationale, including tests that validate it, considering possible rough edges, etc.

simfinite:
Wow, that was quick. I think you are right, I was a bit hasty with pushing this after just rebasing my changes to the latest changes of develop2 branch. I have been using the change like this for quite a while in my own projects, but I see the points you are making. 

For what it's worth, I do think that using CMAKE_MAP_IMPORTED_CONFIG value(s) could be a sensible default/fallback for the build type to query for Conan dependencies. As I see it, users should be able to override this default by setting CONAN_INSTALL_BUILD_CONFIGURATIONS?

memsharded:
> For what it's worth, I do think that using CMAKE_MAP_IMPORTED_CONFIG value(s) could be a sensible default/fallback for the build type to query for Conan dependencies. As I see it, users should be able to override this default by setting CONAN_INSTALL_BUILD_CONFIGURATIONS?

I think, it should be the other way around. The expectation is that if the current build is a ``Release`` build, the dependencies should also be the ones in the same build_type. And if the user wants to define otherwise, it is possible to define it via profiles. Because this is what CMake would do: it will first try to use dependencies with ``Release`` artifacts, and if that configuration is not found, it can fallback to the ``CMAKE_MAP_IMPORTED_CONFIG`` one. 
But with the proposed change, Conan would behave different than CMake, it will ignore the current defined build-type, and it will jump directly to try to get the dependencies defined from the alternate configuration defined in CMAKE_MAP_IMPORTED_CONFIG.

Plus, there is one important thing. With modern CMake (I think >=3.27), and the new ``CMakeConfigDeps`` generation, this might no longer be necessary. CMake can already default CMAKE_MAP_IMPORTED_CONFIG without explicit configuration, and ``CMakeConfigDeps`` define build-types better, so this can naturally happen.

simfinite:
Ok, it seems I was misusing `CMAKE_MAP_IMPORTED_CONFIG` all along. The switch to controlling which configuration to be installed via `conan_provider.cmake` is `CONAN_INSTALL_BUILD_CONFIGURATIONS`, correct? So if I have a custom build type in my CMake-based project that I don't want to expose to Conan when installing dependencies, this variable is what I set in addition to `CMAKE_MAP_IMPORTED_CONFIG_<CUSTOM_BUILD_TYPE>` to have CMake's fallback mechanism choose a compatible binary. Does that make sense?

At the time, I implemented my `CMAKE_MAP_IMPORTED_CONFIG` work-around (let's call it that), there was no support in specifying the configurations to install in `conan_provider.cmake`. Also, I wasn't aware of `CMakeConfigDeps` and will give it a try. Thanks for pointing that out to me.

memsharded:
> The switch to controlling which configuration to be installed via conan_provider.cmake is CONAN_INSTALL_BUILD_CONFIGURATIONS, correct? So if I have a custom build type in my CMake-based project that I don't want to expose to Conan when installing dependencies, this variable is what I set in addition to CMAKE_MAP_IMPORTED_CONFIG_<CUSTOM_BUILD_TYPE> to have CMake's fallback mechanism choose a compatible binary. Does that make sense?

Yes, this makes sense for custom configurations.

Thanks for the feedback!

simfinite:
For the record, I found that setting `CONAN_INSTALL_BUILD_CONFIGURATIONS` is not sufficient for completely hiding custom CMake build types from Conan when using the `conan_provider.cmake` script as it will add the current build type as the project's own/self build config. At this point, I circled back to again implement a work-around to not add the `&:build_type=${CMAKE_BUILD_TYPE}` if `CONAN_INSTALL_BUILD_CONFIGURATIONS` is defined:
```
            set(_self_build_config "")
            if(NOT DEFINED CONAN_INSTALL_BUILD_CONFIGURATIONS AND NOT _multiconfig_generator AND NOT _build_config STREQUAL "${CMAKE_BUILD_TYPE}")
                set(_self_build_config -s &:build_type=${CMAKE_BUILD_TYPE})
            endif()
```

From what I understand, the only other way would be to 
1. Refrain from defining custom build types and rather use some custom flags to alter build types that are well-defined across all tools involved, e.g. use `DEBUG` build with custom flags for coverage instrumentation etc.. 
2. Make the custom build type available/known to Conan

It's a bit of an unfortunate situation when coming from a CMake-first workflow where I only want to use Conan to retrieve and consume a few packages. It's either dropping things that have worked over decades (custom CMake build types) or complicating the workflow having to install custom Conan configurations with no apparent benefit other than satisfying the tooling.

memsharded:
> Refrain from defining custom build types and rather use some custom flags to alter build types that are well-defined across all tools involved, e.g. use DEBUG build with custom flags for coverage instrumentation etc..


This is not intended, it should be possible to use custom build types. Conan has some provisions for it in different parts of the recipes, like for example ``CMakeDeps`` generators have the ``.configuration`` field so it can be defined in recipes: https://docs.conan.io/2/reference/tools/cmake/cmakedeps.html#configuration

> Make the custom build type available/known to Conan

This is also possible via extending settings with ``settings_user.yml``, but might require some extra work defining flags and other stuff.


> It's a bit of an unfortunate situation when coming from a CMake-first workflow where I only want to use Conan to retrieve and consume a few packages. It's either dropping things that have worked over decades (custom CMake build types) or complicating the workflow having to install custom Conan configurations with no apparent benefit other than satisfying the tooling.

But I think you can implement your workflow as you wish, by customizing the ``conan_provider.cmake`` to tailor your needs. What it is not possible is to implement a generic ``conan_provider.cmake`` that works for all users, because it is mostly impossible to know in advance if users will prefer to use dependencies aligned with their current build-type or use other build-type for dependencies, and the current build-type and current CMAKE_MAP_IMPORTED_CONFIG is not enough to know. 
So it is totally fine that you implement your logic in your ``conan_provider.cmake``. The one in this repo doesn't aim to be the only one and the only truth for all users. But adapting it to different use cases is expected and a valid use case.

PR #730, opened 2026-05-12
Make `conan install` command copyable

### LINKED ISSUE #729: Please make the "conan install" message copy-paste able
Hi! This request is purely an aesthetic ask about the cmake logging/messaging, but could you make the cmake status message print the exact conan command that it's running?
Currently, CMake is treating the list of args as a single semicolon-separated strings, but the other args are still space separated, making it confusing at first glance what's happening.
I'm just setting up a conan and cmake integration for my team and so running both cmake and conan commands often. This would let me copy/paste the install command directly from the output messages, to test the conan command manually when needed. Our project's cmake configuration takes a small amount of time and so slows my testing speed when trying to look into issues with our packages.
Here's an example of what the output currently looks like:
`-- CMake-Conan: conan install /Users/jswartw/codebase -of=/Users/jswartw/codebase/build/my_preset/conan --profile:host=android;--profile:build=default;-s;build_type=Debug;--build=missing;-s;arch=armv8`
and what I would like it to look like
`-- CMake-Conan: conan install /Users/jswartw/codebase -of=/Users/jswartw/codebase/build/my_preset/conan --profile:host=android --profile:build=default -s build_type=Debug --build=missing -s arch=armv8`
I think the fix would just be joining the list with spaces before printing, which seems to be done on this line: https://github.com/conan-io/cmake-conan/blob/develop2/conan_provider.cmake#L473
Thanks!

ChrisThrasher:
Closes #729 

When ARGN contains CMake lists, those lists get expanded within quotes. This entails that the semicolons that delimit list items are printed in the log message. To make this log message reflect the actual shell command being ran, we want to strip out those semicolons. We do this by joining each list element with spaces to create a new string.

### DIFF
--- a/conan_provider.cmake
+++ b/conan_provider.cmake
@@ -470,7 +470,8 @@ function(conan_install)
     set(conan_output_folder ${CMAKE_BINARY_DIR}/conan)
     # Invoke "conan install" with the provided arguments
     set(conan_args -of=${conan_output_folder})
-    message(STATUS "CMake-Conan: conan install ${CMAKE_SOURCE_DIR} ${conan_args} ${ARGN}")
+    list(JOIN ARGN " " argn_str)
+    message(STATUS "CMake-Conan: conan install ${CMAKE_SOURCE_DIR} ${conan_args} ${argn_str}")
 
 
     # In case there was not a valid cmake executable in the PATH, we inject the


JakeSwartwoutTrimble:
Just tested and this looks to solve my issue, thanks 👍

JakeSwartwoutTrimble (approved):

memsharded (approved):

memsharded:
Looking good, thanks for the contribution!

PR #733, opened 2026-06-04
Ensure discovered CMake path is appended correctly

JakeSwartwoutTrimble:
When CMake is not already in the path, the directory of the `CMAKE_COMMAND` is stored into `PATH_TO_CMAKE_BIN` to append to the path. Most systems use ":" (colon) as a separator, but on windows the path needs to be separated by ";" (semicolon).

Tested that when CMake isn't in my path, it gets added correctly for:
- Windows Git-Bash terminal configuring windows build -> correctly uses windows ";" path separator
- Windows Android Studio app doing a gradle-sync to configure cross-building for android -> correctly uses windows ";" path separator
- MacOS zsh terminal configuring ios build -> correctly uses the standard ":" path separator

### DIFF
--- a/conan_provider.cmake
+++ b/conan_provider.cmake
@@ -478,7 +478,11 @@ function(conan_install)
     # same we used to invoke the provider to the PATH
     if(DEFINED PATH_TO_CMAKE_BIN)
         set(old_path $ENV{PATH})
-        set(ENV{PATH} "$ENV{PATH}:${PATH_TO_CMAKE_BIN}")
+        if(CMAKE_HOST_WIN32)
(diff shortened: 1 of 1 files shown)


memsharded (approved):
Thanks for the contribution @JakeSwartwoutTrimble 

Looks good to me

memsharded:
It would be necessary to sign the CLA in order to merge it, please check it. Thanks

JakeSwartwoutTrimble:
Thanks! Looks like I had accidentally put my email in wrong so it wasn't getting linked to my CLA signing, looks to be all good now

memsharded:
Excellent, many thanks!

PR #736, opened 2026-09-03
Retry conan install offline if remote comms fails

planetmarshall:
If conan install fails due to an error communicating with a remote, retry using the local cache only. (Checks for an Error from the `HttpsConnectionPool`)

I have found this workaround useful when working offline or when there are upstream issues with the remotes.

### DIFF
--- a/conan_provider.cmake
+++ b/conan_provider.cmake
@@ -485,12 +485,28 @@ function(conan_install)
         endif()
     endif()
 
-    execute_process(COMMAND ${CONAN_COMMAND} install ${CMAKE_SOURCE_DIR} ${conan_args} ${ARGN} --format=json
+    foreach(attempt RANGE 1)
+        execute_process(COMMAND ${CONAN_COMMAND} install ${CMAKE_SOURCE_DIR} ${conan_args} ${ARGN} --format=json
                     RESULT_VARIABLE return_code
                     OUTPUT_VARIABLE conan_stdout
                     ERROR_VARIABLE conan_stderr
                     ECHO_ERROR_VARIABLE    # show the text output regardless
                     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+        if(NOT "${return_code}" STREQUAL "0")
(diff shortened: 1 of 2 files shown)


memsharded (commented):
Hi @planetmarshall 

Thanks for your contribution.

However, I am afraid that this change doesn't seem aligned with the general Conan design. It is not expected that a failure to connect to one remote produces a fallback to using the cache only. While it can look convenient, this silent fallback is something that can produce unexpected behaviors and bugs, and this is why the main Conan cli application does not implement this behavior, and we think that this CMake ``cmake-conan`` integration probably shouldn't do the same.

Users can perfectly define the behavior they wan in their provider, but having this in the upstream ``cmake-conan`` seems too much at the moment. Thanks anyway for the contribution!

planetmarshall:
> 
> It is not expected that a failure to connect to one remote produces a fallback to using the cache only. While it can look convenient, this silent fallback is something that can produce unexpected behaviors and bugs, and this is why the main Conan cli application does not implement this behavior, and we think that this CMake `cmake-conan` integration probably shouldn't do the same.
> 

I would argue that whilst adding `--no-remote` to the arguments when running `conan install` directly is trivial, doing the same when running via CMake is not - particularly when using an IDE. However, as you say it is easy enough to modify `cmake-conan` for my own purposes. Thanks for taking a look.

memsharded:
> I would argue that whilst adding --no-remote to the arguments when running conan install directly is trivial, doing the same when running via CMake is not - particularly when using an IDE. However, as you say it is easy enough to modify cmake-conan for my own purposes. Thanks for taking a look.

Yes, it is easy to do, but that doesn't mean that it is a silent or automatic fallback. It implies a explicit decision of the user to ignore the remotes and install only from the local cache. That exact same decision can be taken by the ``cmake-conan`` user by defining ``CONAN_INSTALL_ARGS`` to contain ``--no-remote`` argument, which might also be defined via IDE as a general CMake variable if desired, so IDE integrations might be able to still opt-in into the decision to ignore the remotes if desired.

